Issue #4, opened 2016-08-29
add/set/etc should return the results of the functions they wrap 


dev-zero:
When calling `cache.add` or `cache.set`, the result is currently lost. This makes it necessary to call `cache.cache.add` instead to get the result of the underlying cache implementation (and to be able to implement a cache-based locking mechanism using `cache.add`).

Since those functions are bare wrappers around a caching implementation (Werkzeug most of the time), I expected them to also return the results.

sh4nks:
Thanks for reporting & sending a PR!

PR #5, opened 2016-08-30
Make wrappers return the wrapped results

dev-zero:
Functions like `add`, `set`, etc usually return a boolean to indicate
whether the operation was successful. Currently, the wrappers throw that
return value away. Even `get_dict` does not return anything even though
it should.

This fixes issue #4.


sh4nks:
Awesome, thanks!

PR #8, opened 2016-10-20
modify tox.ini to use nose instead of unittest2, same as .travis.yml

neilsh:
The tox.ini file was set up to use unittest2 and unittest2py3k, which was kind of weird since unittest2 was really only necessary for python versions < 2.7 (the tox.ini only has versions >= 2.7). I modified the tox.ini to use nose, which is also already included in the requirements.txt and used by the .travis.yml.


sh4nks:
Thanks!

PR #9, opened 2016-11-14
Added support for caching functions under properties

sallyruthstruik:
See https://github.com/sh4nks/flask-caching/issues/16


sh4nks:
Thanks for this PR.

I would appreciate if you could move the tests to the ``test_cache.py`` file for consistency's sake.

Sorry that it took so long.

sallyruthstruik:
Sorry for long reply, 

I've done your requests.  I'm not sure  I've put changes in correct version block, so if I'm wrong please tell me.

sh4nks:
After giving it some thoughts I decided that I'm not merging this. At the moment you just cache the result of a method and return it in a property:
```python
class Foo(object):

    #  ...

    @property
    def foobar(self):
        return self.get_foobar()

    @cache.memoize()
    def get_foobar(self):
        pass
```
If there is demand for this functionality I might reconsider this feature.

huangxiaohen2738:
Ungraced

indiquant:
Hi @sh4nks , I was looking online for a solution like this and I stumbled upon this PR. There is certainly a need for this feature. I would greatly appreciate if you could open this PR for merge.

Cc. @sallyruthstruik

bw984:
I agree with @indiquant that there are use cases where excluding parameters from the key would be very useful.

PR #14, opened 2016-12-07
Change redis import mechanism to avoid ignoring ImportErrors

jkgeyti:
In response to https://github.com/sh4nks/flask-caching/issues/13

### DIFF
--- a/flask_caching/backends.py
+++ b/flask_caching/backends.py
@@ -65,38 +65,37 @@ def filesystem(app, config, args, kwargs):
     kwargs.update(dict(threshold=config['CACHE_THRESHOLD']))
     return FileSystemCache(*args, **kwargs)
 
-# RedisCache is supported since Werkzeug 0.7.
-try:
+def redis(app, config, args, kwargs):
+    # RedisCache is supported since Werkzeug 0.7. 
+    # Place RedisCache import here to not interfere with the
+    # rest of the cache backends for users running Werkzeug <0.7
     from werkzeug.contrib.cache import RedisCache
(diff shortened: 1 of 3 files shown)


sh4nks:
Looks good, the only thing I'd change is to move the RedisCache import to the top of the file (where the other Cache classes are imported. I think it's safe to remove support for version < 0.7 of werkzeug.

Maybe wrap the import in a try-except and raise a exception if redis can't be imported (I have been inspired from [here](https://github.com/pallets/werkzeug/blob/master/werkzeug/contrib/cache.py#L555-L558))

```python
def redis(...):
    try:
        import redis
    except ImportError:
        raise RuntimeError('no redis module found')
```

jkgeyti:
Done. You may also want to consider moving `backends` into its own module, so you can better control what's being exposed through an `__init__.py`.

sh4nks:
Thanks, I really appreciate it! That's actually a neat idea. Do you want to do that?

jkgeyti:
Sure. I've moved backends to a separate folder and added an init file. It's untested, so you might want to take a close look at whether you agree with the solution. You might want to use `__all__` instead.

sh4nks:
Thanks! Looks like adding a ``__all__`` in ``backends.py`` is the way to go and then use ``from .backends import *`` in ``__init__.py``. I'll just add that and release a new version.

sh4nks:
The new version should be on PyPI now :)

johaven:
On last version (1.7.2) no warning is fired if redis module cannot be imported (or missing)

Issue #18, opened 2017-02-02
Function having only *args and *kwargs fails


PonteIneptique:
In the context of my plugin flask-capitains-nemo ( https://github.com/Capitains/flask-capitains-nemo ), I am trying to get memoized a certain list of functions. I currently cache dynamically these one through : 

```python
        if self.cache is not None:
            for func, instance in self.cached:
                setattr(instance, func.__name__, self.cache.memoize()(func))
```

It works well, as long as I don't cache functions that have no enforced args or kwargs schemes but are actually route dispatcher : 
 
```python
def render(self, template, **kwargs):
```

will always create the key : `flask_nemo.Nemo.render('main::collection.html',){}` as the output of self._memoize_kwargs_to_args will not work with a `**kwargs` system. It is most likely the result of the use of `get_arg_names`. which will find only one defined arg.

I'll be trying to propose a fix for this issue.

sh4nks:
> I'll be trying to propose a fix for this issue.

Thanks, could you also add a little test for it?

PonteIneptique:
Well, it would not be a contribution if there wasn't ;)

PonteIneptique:
Just so I know : what timezone are you in ? Just to know where I can expect answer and not :)

sh4nks:
I am in CET. I am off of work today so I have some time to spent on my projects :)

PonteIneptique:
It's good to find someone in CET ! Patch will be coming in the next hour or so

PonteIneptique:
Thanks for the support ! Was really cool to be in sync ;)

sh4nks:
New release is out :)

PR #19, opened 2017-02-02
Fixed issue #18 : when function as **kwargs, keep unknown kwargs for key

PonteIneptique:
I don't know how this would work for *args but at least it works for kwargs ;)
If I find myself in the *args situation and this bugs, I'll propose a second patch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ docs/_build/*
 *.egg-info
 build/*
 dist/
+.idea/
+.idea/*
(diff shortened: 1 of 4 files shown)


sh4nks:
Thanks! Could you also add a little note in the changelog (https://github.com/sh4nks/flask-caching/blob/master/CHANGES)?

PonteIneptique:
Should I bump as well ?

sh4nks:
I'll do that :)

PonteIneptique:
@sh4nks Done :)

sh4nks:
Thanks!

PR #26, opened 2017-03-28
return current_app by default from self.cache

rayctas:
This addresses an issue where an app instance would
look at the wrong cache in a multi-app setup.
The last app passed to `cache.init_app` would stick around
and all the preceding apps would use its cache prefix.

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -219,7 +219,7 @@ class Cache(object):
 
     @property
     def cache(self):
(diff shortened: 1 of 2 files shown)


sh4nks:
Thanks! btw, your PR didn't made the test fail (it was apparently already failing before).

rayctas:
Ok, they did pass locally for me and I was planning to investigate further but didn't get around to it.

PR #29, opened 2017-04-27
Remove useless line in test_22_cached_view_forced_update

patelnav:
Doesn't look like forced_update_fn is doing anything

### DIFF
--- a/test_cache.py
+++ b/test_cache.py
@@ -713,9 +713,6 @@ class CacheTestCase(unittest.TestCase):
     def test_22_cached_view_forced_update(self):
         forced_update = False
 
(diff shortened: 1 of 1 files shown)


sh4nks:
Thanks!

patelnav:
heyo, are there any instructions on what servers need to be started to make the tests work locally? Most of them went along fine, but I didn't have redis installed so a bunch started failing

sh4nks:
For redis you have to make sure that the redis server is running and that you have installed the redis package from pypi. Redis, for example, is configured in those lines: https://github.com/sh4nks/flask-caching/blob/master/test_cache.py#L806-L813

PR #38, opened 2017-06-20
Specify Werkzeug version requirement (fix #37)

### LINKED ISSUE #37: Set werkzeug 0.12 as minimum requirement
The uwsgi backend introduced a dependency on Werkzeug 0.12+.
Without specifying this requirement, the deployment break in existing environment because it does not force the `Werkzeug` dependency upgrade.

noirbizarre:
This PR avoid breakage for environment with Werkzeug<0.12 already installed because the uwsgi backend depends on Werkzeug 0.12+

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,8 @@ setup(
     zip_safe=False,
     platforms='any',
     install_requires=[
-        'Flask'
+        'Flask',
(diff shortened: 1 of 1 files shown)


sh4nks:
Thanks! Will push out a v1.3.1 asap.

PR #41, opened 2017-07-18
try catch on memoize() to match cached()

justinwp:
Might as well have the same behavior for both cached() and memoized().

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -664,18 +664,30 @@ class Cache(object):
                     f, *args, **kwargs
                 )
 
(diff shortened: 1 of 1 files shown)


sh4nks:
I'd prefer it the other way around - adjusting the ``memoize()`` method to match ``cached()``.

justinwp:
that is what this change does, modify `memoize()` to match `cached()`

sh4nks:
Ah sorry.. i mixed it up - what I wanted to say is if you could modify ``cached()`` to match ``memoize()``.

sh4nks:
Added with #60 but thanks anyway!

PR #43, opened 2017-09-12
Support multiple params and use md5 for consistent hashing

johnistan:
This PR fixes a couple of issues I found with the `query_string=True` implementation

- Python 3 now randomizes `hash` on each start up. [Docs](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED) This means that if you reuse the cache across flask deploys or multiple apps deployed behind a load balancer, you will have extra cache misses.
- URL params can be repeated. The current implementation ignored that and built the cache from the first repeated url param

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -398,13 +398,14 @@ class Cache(object):
                 # provided.
                 args_as_sorted_tuple = tuple(
                     sorted(
-                        (pair for pair in request.args.items()),
-                        key=itemgetter(0),
+                        (pair for pair in request.args.items(multi=True))
                     )
                 )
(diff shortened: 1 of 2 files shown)


sh4nks:
Thanks!

YaraslauZhylko:
Hi @sh4nks,
When are you going to release those changes as a separate version on PyPI?
Repeated URL params support is exactly what I need for my API. :)

sh4nks:
Thanks for reminding me! Just pushed 1.3.3 to PyPI ;)

YaraslauZhylko:
That's just awesome! :)
Thank you!

PR #48, opened 2017-10-19
Clarify usage of CONFIG_OPTIONS for various backends

dwlocks:
Werkzeug does not consistently handle options to its cache backends. We
clarify how flask-cache maintains configuration consistency where
possible and provide a workaround example.


sh4nks:
Thanks!

PR #50, opened 2017-11-01
Fix travis coveralls support

neilsh:
This should resolve the `coveralls: command not found` errors at the end of each travis build.


sh4nks:
Thanks!

PR #51, opened 2017-11-17
Improve the syntax and grammar in documentation and docstring

greyli:
Maybe we should remove the details in docstring?


sh4nks:
Thanks!

PR #60, opened 2018-02-28
Memoize Try Except

mlcamilli:
This PR:

- Adds a Try Except to the memoize function that mirrors the functionality that exists in `cache`.  Basically if the application *cannot* reach the cache, it shouldn't error out, but rather return the call seamlessly without using the cache. This was done in the `cache` decorator but was never ported to the memoize function. 

- Also, noticed that when there is an exception, it checks to see if the app has debug enabled and if it does, raises the exception. However calling current app causes an exception itself, as it is not in the application context at that time. I changed these calls to use `self.app` instead which works as the cache is instantiated with the current app. Below is the error I was running into when current_app was called.

```
Traceback (most recent call last):
  File "/src/flask-caching/flask_caching/__init__.py", line 679, in decorated_function
    if current_app.debug:
  File "/usr/local/lib/python3.4/site-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/usr/local/lib/python3.4/site-packages/werkzeug/local.py", line 306, in _get_current_object
    return self.__local()
  File "/usr/local/lib/python3.4/site-packages/flask/globals.py", line 34, in _find_app
    raise RuntimeError('working outside of application context')
RuntimeError: working outside of application context
```

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -666,10 +666,10 @@ class Cache(object):
                 if self._bypass_cache(unless, f, *args, **kwargs):
                     return f(*args, **kwargs)
 
-                cache_key = decorated_function.make_cache_key(
-                    f, *args, **kwargs
-                )
                 try:
+                    cache_key = decorated_function.make_cache_key(
+                        f, *args, **kwargs
+                    )
 
                     if callable(forced_update) and forced_update() is True:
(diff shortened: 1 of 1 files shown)


sh4nks:
Cool, thanks!

PR #69, opened 2018-04-12
Specify Pypi package name in ReadMe.

rangerscience:
It's not obvious if you're unfamiliar with python import vs package naming conventions.


sh4nks:
Thanks for your PR but I already did that in this commit: 4611cbddf8cf03af45e4bdadb8a9a75991cdbe9e

PR #74, opened 2018-06-29
Include tests in PyPI tarball

dotlambda:


alanhamlett:
No, we shouldn't include tests in releases.

dotlambda:
Some of the big names are doing it: Numpy, Cython, ...
I think it makes sense because e.g. distributions might want to include a test phase in their build instructions, just to make sure the package works fine with the packaged versions of all of its dependencies.

alanhamlett:
There are ways around that without having to include tests in releases.

Also, if a release fails to build on a specific environment then a test should be written that correctly mocks that environment and that test should fail in automated/dev environments. That way problems are caught before they reach a release.

There's always a better alternative rather than including unit tests or other dev-only code in releases.

sh4nks:
> Also, if a release fails to build on a specific environment then a test should be written that correctly mocks that environment and that test should fail in automated/dev environments. That way problems are caught before they reach a release.

Yep, I agree.

@dotlambda Thanks for your PR nevertheless!

Issue #72, opened 2018-04-26
Deprecate werkzeug.contrib.cache


lepture:
We are going to remove `werkzeug.contrib.cache` from werkzeug. Flask-Caching is depending on it, can you copy that part into your repo and maintain it in Flask-Caching.

https://github.com/pallets/werkzeug/issues/1249

neilsh:
For reference, here's the code for `werkzeug.contrib.cache`: https://github.com/pallets/werkzeug/blob/master/werkzeug/contrib/cache.py

Flask-Caching relies on this module for the implementations of several of its caching backends:

https://github.com/sh4nks/flask-caching/blob/039d8aef5e3ebbb70c0f679f4e4452f80f594dd1/flask_caching/backends/backends.py#L11

dchevell:
@lepture is this the sort of approach you're talking about: https://github.com/dchevell/flask-caching

I copied over the code from werkzeug.contrib.cache, replaced all imports in the core library and in tests, and updated the documentation to reference the internal implementations. Werkzeug origins are still clear.

All tests pass, and I tried the modded version in a project of mine that uses Flask-Caching and found no unexpected behaviours.

I'm not 100% confident in opening a PR without confirming whether this approach is the one you're looking for, or one that @sh4nks is happy with.

lepture:
Yes, you can start with a copy of werkzeug.contrib.cache. But I do believe the cache module could be far more better than the one in werkzeug.

sh4nks:
@dchevell that would be ok for now I think!


I actually had in mind to rewrite this extension to be a *simple* wrapper around dogpile.cache which would make maintaining this extension much easier. At the moment this is just an idea.. which lacks motivation and time :/

sh4nks:
I am closing this now as we don't depend on werkzeug.contrib.cache anymore. Yay! :)

PR #75, opened 2018-07-09
Deprecate werkzeug.contrib.cache

dchevell:
As per #72 , this PR simply removes the direct dependency on werkzeug.contrib.cache. As a first step, the werkzeug implementation has been adopted wholesale, and the rest of this library simply updated to use the correct import paths and to reflect that in the docs.

Going forward we may want to remove these and replace them with something similar, but as an immediate change to guard against werkzeug.contrib.cache's planned deprecation this PR makes no attempt to change or improve the cache backends themselves.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -408,7 +408,7 @@ Relevant configuration values
 - CACHE_MEMCACHED_SERVERS
 
 
-.. note:: Flask-Cached does not pass additional configuration options
+.. note:: Flask-Caching does not pass additional configuration options
    to memcached backends. To add additional configuration to these caches,
    directly set the configuration options on the object after instantiation::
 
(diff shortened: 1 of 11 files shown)


dchevell:
Adding werkzeug's test for caching backends

dchevell:
Apparently pytest deciced to [drop support for python3.3](https://github.com/pyinstaller/pyinstaller/pull/3060) as it is now EOL. This breaks some tests running on py3.3 so I've removed that, and added py37-dev to make sure we're testing the latest release.

docs/index.rst
```diff
@@ -409,7 +408,7 @@ Relevant configuration values
 - CACHE_MEMCACHED_SERVERS
 
 
-.. note:: Werkzeug does not pass additional configuration options
+.. note:: Flask-Cached does not pass additional configuration options
```
sh4nks: Flask-Caching

dchevell: I wear the cone of shame! I'll fix it now.

dchevell:
Fixed @sh4nks

sh4nks:
Thanks a lot @dchevell!

PR #77, opened 2018-07-20
parameterize the hash function so alternatives can be used

### LINKED ISSUE #65: Support for sha256 in memoize key for FIPS compliance
I am in the process of migrating my server to a FIPS-compliant OS which doesn't have md5 available which results in a segmentation fault when using the memoize decorator.

I'm wondering if it's possible to override the make_cache_key method or some other way to use sha256 instead of md5 to create the cache key with memoize.

https://github.com/sh4nks/flask-caching/blob/master/flask_caching/__init__.py#L513

Thanks!

buckley-w-david:
This PR resolves #65 by making the hashing function used a parameter in all functions/methods that used to be hard-coded to use md5.

This PR makes a change to all the tests that use affected methods to try with both md5 and sha256, but this might be overkill, tell me what you think @sh4nks and I'll take them out if you'd rather just an individual test or two to verify that multiple hash functions work.

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -339,6 +339,9 @@ class Cache(object):
                              were passed in a different order. See
                              _make_cache_key_query_string() for more
                              details.
+        :param hash_method: Default hashlib.md5. The hash method used to
+                            generate the keys for cached results.
+
         """
         def decorator(f):
             @functools.wraps(f)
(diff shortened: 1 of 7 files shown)


buckley-w-david:
I've refactored the tests to explicitly test different hash functions for each place it is not configurable.

To make it a little cleaner, I also added a `SUPPORTED_HASH_FUNCTIONS` list to the `__init__.py`, wanted to get your input on that.

The ones I picked were ones that were compatible with the usage pattern used previously when it was hardcoded to use `md5`, that also had a explicit constructor in hashlib. Nothing is stopping other methods from working, but I figured that was a good list to test with and explicitly support.

buckley-w-david:
@sh4nks What do you think of the changes?

sh4nks:
One more thing - could you update the docstring to include the hash_method arg? The updated changes did somehow slip through.. sorry

buckley-w-david:
@sh4nks done :+1:

sh4nks:
thanks!

Issue #40, opened 2017-07-07
Flask Caching With Sentinel 


JLiu1272:
I am trying to enable Flask Cache. We are currently using redis-sentinel to handle failover issues, and to increase concurrency performance. However, it seems this way of caching is more difficult to resolve than imagine. Perhaps, you guys can help me. 

```
from flask_caching import Cache
from redis import StrictRedis
from werkzeug.contrib.cache import BaseCache
from redis.sentinel import Sentinel
from app import Config

class SentinelCache(BaseCache):
   def __init__(self, key_prefix=None, **kwargs):
        BaseCache.__init__(self, default_timeout=300)
        host = str(kwargs['host'])
        port = str(kwargs['port'])
        #print(host + ":" + port)
        print(Sentinel([{host:host, port:port}]))
        #Setting self client to be a sentinel 
        self._client = Sentinel([{host:host, port:port}])

def sentinel(app, config, args, kwargs):
  #Get master 
   sentinel = Sentinel(list(map(lambda x: (x.split(':')[0], int(x.split(':')[1])), Config.get('REDIS_SENTINEL_SERVERS').split(','))))
   master = sentinel.discover_master('rdwhizz')
   config['SENTINEL_SERVERS'] = master
   kwargs.update(dict(
      host=master[0],
      port=master[1]
   ))
   print(kwargs)
   return SentinelCache(*args, **kwargs)

cache = Cache(config={'CACHE_TYPE':'app.cache.sentinel'})
```

Here is what I have currently. We were able to retrieve the master host and port. But what we are  are unable to do is sub class Base Cache to enable Sentinel Caching.  

Thank you!

sh4nks:
Sorry for my late reply.

Did you solve the problem? Can I still help you somehow?

sh4nks:
Checkout this branch for basic redis-sentinel support: https://github.com/sh4nks/flask-caching/tree/redis-sentinel

Things to TODO/investigate:
- What happens when you want to use 2 master? Is this even possible?
- Add documentation about the new backend
- Testing with more than one sentinel

This is the config I used to test sentinel with:

```
sentinel monitor mymaster 127.0.0.1 6379 2
sentinel down-after-milliseconds mymaster 60000
sentinel failover-timeout mymaster 180000
sentinel parallel-syncs mymaster 1
```

JLiu1272:
Hi yes, I figured it out. But thank you so much for adding that. I am sure it will be useful for some others as well.

[...]

PR #90, opened 2019-02-14
WIP: Redis sentinel support

gergelypolonkai:
Documented solution for #40 

This is basically the original commit of @sh4nks (rebased onto the current master) with some documentation added. It still needs some testing and possibly some additions, but it seems to be a good start.


gergelypolonkai:
While live testing this, i found there are a lot of things missing; for example there is no way to authenticate with the sentinel instances. I’ll add this soon to this PR.

sh4nks:
This is looking good already! Let me know when it's ready to be merged.

gergelypolonkai:
It would be awesome if someone could give it a go (other than me, i mean). Other than that, it’s pretty much done except some tests maybe.

sh4nks:
I'll try it out as I do have some spare time this afternoon :)

However, I am not exactly sure how to set up Redis-Sentinel - any tips?

gergelypolonkai:
Last time i used [this](https://seanmcgary.com/posts/how-to-build-a-fault-tolerant-redis-cluster-with-sentinel/) article to do it. For testing purposes it’s enough to set up one sentinel with two slaves. If you want to go for 100% (ie. production), you might need at least three sentinels, though.

sh4nks:
I do have some fixes for your PR. Either you apply them or you could also set the PR to allow commits from the maintainers (https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

gergelypolonkai:
The allow maintainers to edit checkbox is ticked, so i guess you should be able to add commits.

I think the refactoring/import reorganising related stuff should be separate from this PR, but that might be just my OCD speaking :)

The `BaseCache.__init__()` call might be justified, as `RedisCache.__init__()` does things we don’t want `RedisSentinelCache.__init__()` to do. Namely checking the `host` argument, which is not passed to `RedisSentinelCache` at all (it needs `sentinel_hosts` instead). The current soluotion, however, makes PyLint to yell at me for not calling the constructor of the parent class, and that makes me feel uncomfortable. But unless we want to completely duplicate the methods of `RedisCache`, we might instead create a `RedisCacheBase` class that might be the copy of the current `RedisCache` without `__init__`, then recreate `RedisCache` with its current `__init__`, and keep the `RedisSentinelCache` as is (with the new base class). Another option might be to completely ditch `RedisSentinelCache` and modify `RedisCache` so it can connect to a sentinel setup. That, however, might be slightly harder work.

You renaming `self._read_clients` to `self._read_client` is valid from a grammatic PoV, but might be wrong otherwise. While `_write_client` will always be one specific client (as told by the sentinels), `_read_client` is a connection pool with one or more slaves. I’m still OK with renaming it, though, i’m just saying :)

sh4nks:
> The `BaseCache.__init__()` call might be justified, as `RedisCache.__init__()` does things we don’t want `RedisSentinelCache.__init__()` to do.

But it does not work with having `BaseCache.__init__()`. I did the tests using the example app in `examples/`.

Ok, then lets just exclude the last patch ([PATCH 3/3] Use super()-style for initializing inherited parent) from this PR.

My bad, I actually just wanted to pass the same arguments as were passed before.

> You renaming `self._read_clients` to `self._read_client` is valid from a grammatic PoV, but might be wrong otherwise. While `_write_client` will always be one specific client (as told by the sentinels), `_read_client` is a connection pool with one or more slaves. I’m still OK with renaming it, though, i’m just saying :)

I was just doing the fastest way of fixing the things - because in RedisCache, everywhere `self._read_client` is used instead of `self._read_clients`.

gergelypolonkai:
Oh, i take a day out and you merged it already? 😃 Thanks a lot!

Do you think you will release it any time soon?

sh4nks:
Yeah I'll push out a new release soon. I still have a couple of changes in the pipeline that I would like to have in the release.

PR #94, opened 2019-03-05
Fix delete_many method stop at first error

### LINKED ISSUE #64: delete_many stop at first error
```python
    def delete_many(self, *keys):
        """Deletes multiple keys at once.

        :param keys: The function accepts multiple keys as positional
                     arguments.
        :returns: Whether all given keys have been deleted.
        :rtype: boolean
        """
        return all(self.delete(key) for key in keys)
```
Currently the delete loop stops at the first item that is False (it's the behavior of the **all** function if evaluation is done inside)
This is annoying because you have to check the presence in the cache of all the keys you want to delete.
The default behaviour should be to continue the deletion despite the errors.
In this way the trick is to bypass the evaluation during the loop:
```python
 return all([self.delete(key) for key in keys])
```
Or add this as option ?
```python
def delete_many(self, *keys, ignore_errors=False)
```
Note: I found this problem with the file system cache

sh4nks:
Fixes #64

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,18 @@
 Changelog
 ---------
 
+Version 1.6.0 Unreleased
+````````````````````````
+
+- The ``delete_many`` function is now able to ignore any errors and continue
+  deleting the cache. However, in order to preserve backwards compatibility,
+  the default mode is to abort the deletion process. In order to use the new
+  deletion mode, one has to flip the config setting ``CACHE_IGNORE_ERRORS`` to
+  ``True``. This was and still is only relevant for the **filesystem** and
+  **simple** cache backends.
+  PR `#94 <https://github.com/sh4nks/flask-caching/pull/94>`_.
+
+
 Version 1.5.0 2019-02-23
 ````````````````````````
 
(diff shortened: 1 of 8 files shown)


sh4nks:
@buckley-w-david This is the least invasive way I was able to come up with. What do you think about the fix?

buckley-w-david:
Looks alright to me.

Technically
https://github.com/sh4nks/flask-caching/blob/f4bf4696986c32ce7bc44a47f50266c0fa3f430a/flask_caching/backends/base.py#L136
is slightly inefficient since it will traverse the collection twice, once to actually perform the deletions and again to evaluate `all`. If you actually care about that, instead of doing the list comprehension instead you could just do a for loop over the collection and keep track of the result as you go (so doing an `and` with the current value and the result of the delete each iteration).

I don't really think it matters very much however.

What's more interesting to me is that fact that the change failed the `test_backend_cache.py::TestMemcachedCache::test_generic_timeout` test in the python-3.7 environment. It's possible that the change actually is failing the test on that version of python, but what seems more likely is that it's a flaky test.

sh4nks:
Thanks for the fast feedback! Ah right, we could still change it later if performance gets a problem :)

I have no idea why the tests are failing on Python 3.7 because I was developing and testing the whole thing on 3.7.2.
```
  py27: commands succeeded
ERROR:  py35: InterpreterNotFound: python3.5
ERROR:  py36: InterpreterNotFound: python3.6
  py37-dev: commands succeeded
```

Issue #91, opened 2019-02-18
Disable caching based on the response http status code


ronlut:
Hi.
I'm looking for an option to disable caching for certain response status codes.
My flask endpoint queries another endpoint, and I return different status codes (in the 4XX-5XX range) depending on the problem I encountered with the second endpoint.
I want to disable caching for responses that are not within the "valid" range (2XX-3XX).
Is that somehow possible with flask-caching? The `unless` function is evaluated before my view function is invoked so I can't use the response.

Thanks

sh4nks:
If this is not achievable with ``unless`` you have to implement your solution to disable caching for your specific issue. Sorry.

PR #99, opened 2019-03-15
Control caching based on response

volodymyrss:
I might be missing entierly the intent of the developers, and possibly there is a much better solution to my problem (#91). 
This is just what I implemented which is sufficient for my needs, I wonder if it makes any sense.

### DIFF
--- a/tests/test_view.py
+++ b/tests/test_view.py
@@ -56,15 +56,18 @@ def test_cached_view_unless(app, cache):
     assert the_time == rv.data.decode('utf-8')
 
 def test_cached_view_response_filter(app, cache):
-    app.return_code = 500
 
     @app.route('/a')
     @cache.cached(5, response_filter=lambda x: x[1]<400)
     def cached_view():
         print("app.rt",app.return_code)
         return (str(time.time()), app.return_code)
-
+    
     tc = app.test_client()
+    
+    # 500 response does not cache
+
+    app.return_code = 500
 
     rv = tc.get('/a')
     the_time = rv.data.decode('utf-8')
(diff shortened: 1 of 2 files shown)


sh4nks:
This looks good to me from a first look! I'll take a closer look when I am home from work.

tests/test_view.py
```diff
@@ -55,6 +55,41 @@ def cached_view():
 
     assert the_time == rv.data.decode('utf-8')
 
+def test_cached_view_response_filter(app, cache):
+
+    @app.route('/a')
+    @cache.cached(5, response_filter=lambda x: x[1]<400)
+    def cached_view():
+        print("app.rt",app.return_code)
```
sh4nks:
```suggestion
```
Please delete the print statement

volodymyrss:
thanks for the suggestion!
I see you do not use print even in the test, I should have followed that. Could I ask why, for educational purposes?

sh4nks:
Wait, you were able to commit my suggested changes directly from github? 😄 

Generally speaking, instead of using print statements, the information should be logged using a dedicated logger. The logger can then be configured to only print certain log levels to the console, or log everything to a file.

In our case, by reading the test cases you have written one can easily guess what the result is -- and in case the tests fail, pytest will show us all the needed information anyway.

Hope that clears things up!

volodymyrss:
Ok, I see, thanks!
I would usually use logger except for the tests, but then no reason to treat tests as second-class code. 

Yeah, the technology of github is growing beyond our imagination. There was just a button to accept the suggestion, first time I see such a feature.

PR #100, opened 2019-03-24
Can't use UWSGICache in Python 2.7

pogliamarci:
Hello,

Using Python 2.7 and UWSGICache to add a caching layer to a legacy Flask application, I get AttributeErrors when uwsgi.py attempts to call any function from the uwsgi module.

Apparently, the `import uwsgi` is importing the `uwsg.py` itself instead of the top-level uwsgi module.

As a solution, this pull request renames `uwsgi.py` in `uwsgicache.py` to avoid the name clash, mostly in the same way it's already done for the redis backend (judging from the TODO comment at line 17 of `backends/__init__.py`).


sh4nks:
LGTM! I had the same issue with redis..

PR #101, opened 2019-04-02
Fix: Introspect python3 functions properly using varkw

shakeelsoogun:
The `getfullargspec` function appears to return different values than that returned by `getargspec` (`keyword` in `getargspec` has changed to `varkw`), so actually attempting to provide an `unless` function with arguments to the `cached` decorator fails on python3.

In addition, the previous condition only allowed arguments to be passed through if the `unless` function took in at least one argument and variable positional and keyword args which seemed a little restrictive (what if the original only used kwargs?). Unless there is a good reason for doing this that I'm not aware of?

Let me know what you think about this @sh4nks !

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -660,11 +660,21 @@ class Cache(object):
             try:
                 # Python >= 3.0
                 argspec = inspect.getfullargspec(unless)
+                has_args = (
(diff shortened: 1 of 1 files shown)


sh4nks:
Thanks! I don't think that there was a reason for that.

PR #108, opened 2019-05-01
Don’t run cached/memoized function if the cached return value is None

### LINKED ISSUE #106: cached and memoize run the cached function if the cached value is None
Consider the following (silly) scenario:

```
@cache.memoize()
def very_lengthy_operation(param):
    return do_length_stuff(param)
```

Here, `None` might be a completely valid return value, but it takes a long time to derive it from `param`, hence the caching. The second time i call this function, `memoize` will look up this value in the cache and will get `None` from `cache.get()`. Then it [checks](https://github.com/sh4nks/flask-caching/blob/master/flask_caching/__init__.py#L405) if the value is `None` and if it is, it runs `very_length_operation()` again.

This is because most (all?) cache backends return `None` for any key that is not found in the cache.

gergelypolonkai:
Fixes #106

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -392,8 +392,16 @@ class Cache(object):
 
                     if callable(forced_update) and forced_update() is True:
                         rv = None
+                        found = False
                     else:
                         rv = self.cache.get(cache_key)
+                        found = True
+
+                        # If the value returned by cache.get() is None, it
+                        # might be because the key is not found in the cache
+                        # or because the cached value is actually None
+                        if rv is None:
+                            found = self.cache.has(cache_key)
                 except Exception:
                     if self.app.debug:
                         raise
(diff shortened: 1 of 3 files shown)


sh4nks:
Thanks! Could you fix the failing PyPy and Python 3 tests?

gergelypolonkai:
There you have it, i could even manage not to add `mock` as a dependency.

Also, do you think you can address my open questions?

sh4nks:
Ah, you meant the code comment. Maybe something like this could test the ``None`` return value:
```python
@cache.memoize()
def memoize_none(param):
    call_counter[param] += 1

    if call_counter == 1:
        return None

    return call_counter[param]

# should return "None"
rv_none = memoize_none(1)
assert rv_none is None

cache.delete_memoized(memoize_none)

# should return "2"
rv_none = memoize_none(1)
assert rv_none == 2
```
I haven't tested it but it should theoretically test if its returning ``None``.

PR #109, opened 2019-05-01
RedisCache now supports function as a key_prefix

alexdong:
It's quite useful to use a function, instead of a string, as the `key_prefix`. This is especially useful when we want the cache to be unique based on GET query parameters. For more details, please see https://stackoverflow.com/questions/14228985/what-does-key-prefix-do-for-flask-cache.

### DIFF
--- a/tests/test_backend_cache.py
+++ b/tests/test_backend_cache.py
@@ -226,14 +226,14 @@ class TestRedisCache(GenericCacheTests):
         else:
             host = redis.Redis()
 
-        c = backends.RedisCache(host=host, key_prefix="werkzeug-test-case:")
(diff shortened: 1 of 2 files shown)


alexdong:
I can see that Travis is complaining about failed tests for pypy but those are `memcached` related.

tests/test_backend_cache.py
```diff
@@ -226,14 +226,14 @@ def make_cache(self, request):
         else:
             host = redis.Redis()
 
-        c = backends.RedisCache(host=host, key_prefix="werkzeug-test-case:")
+        c = backends.RedisCache(host=host, key_prefix=lambda: "werkzeug-test-case:")
```
sh4nks: Please add another test case which tests the key_prefix as a function instead of changing this test case.

PR #115, opened 2019-05-25
Make the forced_update of @cached and @memoize accept the same parameters as the decorated function

### LINKED ISSUE #112: forced_update parameter of cached and memoized might receive the some parameters as the function
Maybe i just canʼt see some obvious way, but how the `forced_update` callable is supposed to decide if the value should be refreshed in the cache?

My idea was to pass a `refresh_cache=True` parameter to my function and use that in this callable. However, neither `@memoize` nor `@cached` pass any arguments to `forced_update`.

Is this intentional? I know it would be a breaking change to change this behaviour, though.

gergelypolonkai:
Fixes #112

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -41,6 +41,15 @@ delchars = "".join(c for c in map(chr, range(256)) if c not in valid_chars)
 null_control = (dict((k, None) for k in delchars),)
 
 
+def wants_args(f):
+    """Check if the function wants any arguments
+    """
+
+    argspec = inspect.getfullargspec(f)
+
+    return bool(argspec.args or argspec.varargs or argspec.varkw)
+
+
 def get_arg_names(f):
     """Return arguments of function
 
@@ -364,7 +373,12 @@ class Cache(object):
                             args, kwargs, use_request=True
                         )
 
-                    if callable(forced_update) and forced_update() is True:
+                    if callable(forced_update) and \
+                       (
+                           forced_update(*args, **kwargs)
+                           if wants_args(forced_update)
+                           else forced_update()
+                       ) is True:
                         rv = None
                         found = False
                     else:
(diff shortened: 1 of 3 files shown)


flask_caching/__init__.py
```diff
@@ -46,6 +46,29 @@
     null_control = (dict((k, None) for k in delchars),)
 
 
+def wants_args(f):
+    """Check if the function wants any arguments
+    """
+
+    try:
```
sh4nks: Now that Python 2 is removed, I guess we can simplify this now?

flask_caching/__init__.py
```diff
@@ -390,7 +413,7 @@ def decorated_function(*args, **kwargs):
                             args, kwargs, use_request=True
                         )
 
-                    if callable(forced_update) and forced_update() is True:
+                    if callable(forced_update) and (forced_update(*args, **kwargs) if wants_args(forced_update) else forced_update()) is True:
```
sh4nks: The line is a bit to long. I'd like to keep the line length to 80.

I am using [black](https://github.com/python/black) to keep my python files nicely formatted (we gotta add this to our not yet existing CONTRIBUTING file :P)

gergelypolonkai:
I think i fixed everything we talked about. Also, finally we have a coverage increase! :smile:

tests/test_memoize.py
```diff
@@ -460,6 +460,67 @@ def big_foo(a, b):
...
+
+        def forced_update_func(a, b):
+            call_counter[1] += 1
+            call_params[call_counter[1] - 1] = (a, b)
+
+            return forced_update
+
+        @cache.memoize(5, forced_update=forced_update_func)
+        def memoized_func(a, b):
+            print('memoizing')
```
sh4nks: Is this print statement on purpose?

gergelypolonkai: No, it wasn’t :blush:

PR #116, opened 2019-05-25
Remove Python 2 support

gergelypolonkai:


tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27, py35, py36, py37-dev
+envlist = py35, py36, py37-dev
```
sh4nks: I know it is not entirely relevant for the PR but while you are at could you change the py37-dev environment to py37?

flask_caching/_compat.py
```diff
@@ -12,52 +12,28 @@
...
-        if x is None or isinstance(x, str):
-            return x
+range_type = range
+iterkeys = lambda d: iter(d.keys())
+itervalues = lambda d: iter(d.values())
+iteritems = lambda d: iter(d.items())
+
+text_type = str
+string_types = (str, )
+integer_types = (int, )
```
sh4nks: I think we should remove the lines 18-25 too.

tox.ini
gergelypolonkai: I tried, but apparently itʼnot supported by Travis yet. I havenʼt checked Tox, though.

flask_caching/_compat.py
gergelypolonkai: Yeah, i was about to ask if the compat file should stay at all. I guess iʼll just remove it, then.

tox.ini
gergelypolonkai: Tox does support it. I changed `tox.ini` and will keep my eyes on Travis to see when they drop the suffix, too.

gergelypolonkai:
I fixed what you asked for, and the tests still run fine. the coverage change looks strange, but looking it the numbers it didn’t actually change (the same amount of code lines are covered as before, we just lost some previously tested code.

flask_caching/backends/base.py
```diff
@@ -13,7 +10,7 @@ def iteritems_wrapper(mappingorseq):
 
     """
     if hasattr(mappingorseq, "items"):
-        return iteritems(mappingorseq)
+        return iter(mappingorseq.items())
```
sh4nks: In Python 3, items() returns an iterator and not a list --> ``iter`` is not needed here

flask_caching/backends/memcache.py
```diff
@@ -97,7 +97,7 @@ def get_dict(self, *keys):
         d = rv = self._client.get_multi(_keys)
         if have_encoded_keys or self.key_prefix:
             rv = {}
-            for key, value in iteritems(d):
+            for key, value in iter(d.items()):
```
sh4nks: ``iter(d.items())`` --> ``d.items()``

flask_caching/backends/base.py
gergelypolonkai: Yes, that was suspicious for me, i just couldn’t wrap my head around it early in the morning i guess.

gergelypolonkai:
There, everything is fixed. Take a look and merge when you see fit.

sh4nks:
Everything looks good now! Thanks!

PR #123, opened 2019-06-24
Allow to use __caching_id__ rather than __repr__ as an object caching key

jd:
This allows users to implement __caching_id__ as a class method to override the
default `repr` usage that does not suit all uses cases.

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -71,6 +71,14 @@ def get_arg_default(f, position):
     return arg_def if arg_def != inspect.Parameter.empty else None
 
 
+def get_id_func(obj):
+    return getattr(obj, "__caching_id__", repr)
+
+
+def get_id(obj):
+    return get_id_func(obj)(obj)
+
+
 def function_namespace(f, args=None):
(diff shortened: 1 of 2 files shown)


flask_caching/__init__.py
```diff
@@ -71,6 +71,14 @@ def get_arg_default(f, position):
     return arg_def if arg_def != inspect.Parameter.empty else None
 
 
+def get_id_func(obj):
```
gergelypolonkai: Why this separate function is needed here? As i see you only use this value within `get_id`.

jd: For future reuse and ease readability?

gergelypolonkai: I simply canʼt think of a use case when you need the id generating function but not the ID. And this way you add an unnecessary function call to a oneliner with no significant improvement on readability. How about something like this?

```
def get_id(obj):
    id_func = getattr(obj, '__caching_id__', repr)

    return id_func(obj)
```

jd: 🤷‍♂

PR #127, opened 2019-07-02
add `unlink` to delete bigger key in redis-cache

huangxiaohen2738:


flask_caching/__init__.py
```diff
@@ -264,6 +264,13 @@ def get_dict(self, *args, **kwargs):
         """Proxy function for internal cache object."""
         return self.cache.get_dict(*args, **kwargs)
 
+    def unlink(self, *args, **kwargs):
+        """Proxy function for internal cache object
+        only support Redis
+        """
+        unlink = getattr(self.cache, "unlink", None)
+        return unlink(*args, **kwargs) if unlink is not None else None
```
gergelypolonkai: Since `unlink` is only supported by Redis, other backends (and so will older Redis versions) will default to a no-op here. You should either resort to a plain `delete()` operation if `unlink` is not available, or at least issue a warning that the backend doesn’t support this call.

flask_caching/backends/rediscache.py
```diff
@@ -164,6 +164,17 @@ def inc(self, key, delta=1):
...
 
+    def unlink(self, *keys):
+        """when redis-py >= 3.0.0 and redis > 4, support this operation
+        """
+        if not keys:
+            return
+        if self.key_prefix:
+            keys = [self.key_prefix + key for key in keys]
+        unlink = getattr(self._write_client, "unlink", None)
+        if unlink is not None and callable(unlink):
```
gergelypolonkai: Is there really a case where `unlink` is not `None` here and is not a callable?

flask_caching/__init__.py
huangxiaohen2738: yep. I had the change which you said, but I'm not sure  it's good.

flask_caching/backends/rediscache.py
huangxiaohen2738: I just want to confirm that he does exist.

huangxiaohen2738:
Maybe use the `unlink` in the `clear`

huangxiaohen2738:
@gergelypolonkai Is it ok?

gergelypolonkai:
Yes! Sorry, i completely forgot about this one…

PR #131, opened 2019-07-24
Really skip xprocess if the module is not there

eLvErDe:
Currently it crashes with ImportError


gergelypolonkai:
Sorry, i completely forgot about this. Merging it now, thanks for the fix!

PR #138, opened 2019-10-18
add `enabled` in the config to control apis which were cached

huangxiaohen2738:
1.  Turn off cache conveniently, if use the more cached apis,
If you have the multi instances, you can turn off the cache which is in any instance

2. Use the dict to replace OrderedDict in python3
Less memory and More quickly

@gergelypolonkai

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -16,7 +16,6 @@ import logging
 import string
 import uuid
 import warnings
-from collections import OrderedDict
 
 from flask import current_app, request, url_for
 from werkzeug.utils import import_string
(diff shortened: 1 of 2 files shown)


huangxiaohen2738:
Use the `dict` to replace `OrderedDict` in python3
It uses the less memory and more quickly

sh4nks:
If you don't want to enable the cache on a instance you could simple change the cache type to ``null`` which in my opinion would be better.

I'd rather stay with the ``OrderedDict`` implementation instead of the ``dict`` due to the fact that the order of the ``dict`` is implementation specific.

PR #140, opened 2019-11-12
cache_none option for memoize to deal with concurrency problem

### LINKED ISSUE #135: Concurrency problem
https://github.com/sh4nks/flask-caching/blob/50bc8af30ae9f85714a119c323e8f09100d7a937/flask_caching/__init__.py#L385-L392

If the cache has expired, after the execution of the code "rv = self.cache.get(cache_key)", there may be other processes that have already written values, but here will return a None instead of calling the function to get the new return value. Sorry for the bad English, all of the above are google translation.

joeportela:
Fixes #135 by allowing the user to specify if their function never returns None.  If so, we can safely assume a None returned from cache.get() means the key does not exist.

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -753,6 +753,10 @@ class Cache(object):
                               renewal of cached functions.
         :param hash_method: Default hashlib.md5. The hash method used to
                             generate the keys for cached results.
+        :param cache_none: Default False. If set to True, add a key exists
+                           check when cache.get returns None. This will likely
+                           lead to wrongly returned None values in concurrent
+                           situations and is not recommended to use.
 
         .. versionadded:: 0.5
             params ``make_name``, ``unless``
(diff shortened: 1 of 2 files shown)


joeportela:
Also want to consider the defaulting the value to False or making the kwarg 'cache_none' to make things a bit more clear.

gergelypolonkai:
Yes, please change it to False by default, so we revert back to 1.7.1 functionality. It was much better for everyone, or so it seems.

joeportela:
> Yes, please change it to False by default, so we revert back to 1.7.1 functionality. It was much better for everyone, or so it seems.

Changed the default value to False and also changed the kwarg to 'cache_none'.  I think for a user, the decision to cache None values is a simpler one to make (don't) than deciding if their function may ever return a None value.

gergelypolonkai:
Thanks!

PR #141, opened 2019-11-20
default cached to not cache None

joeportela:
Secondary addition to #135 fix.  Original PR only included fix for memoize and not the cached function.


sh4nks:
@joeportela Why did you remove the ``unlink`` function from the redis backend?

joeportela:
@sh4nks I noticed it was defined twice, seemingly from the resolution of a merge conflict or something. Removing it allowed my test coverage check to pass by reducing the denominator.

sh4nks:
Ahh I see. Thanks for the PR!

PR #147, opened 2019-12-06
use _pickle to replace cPickle in python3

huangxiaohen2738:
And pypy3 link is available.

### DIFF
--- a/flask_caching/backends/filesystem.py
+++ b/flask_caching/backends/filesystem.py
@@ -14,14 +14,13 @@ import hashlib
 import os
 import tempfile
 from time import time
-
 from werkzeug.posixemulation import rename
 
 from flask_caching.backends.base import BaseCache
 
 try:
-    import cPickle as pickle
-except ImportError:  # pragma: no cover
+    import _pickle as pickle
+except ImportError:
     import pickle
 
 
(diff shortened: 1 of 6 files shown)


flask_caching/backends/filesystem.py
```diff
@@ -184,7 +183,7 @@ def set(self, key, value, timeout=None, mgmt_element=False):
             )
             with os.fdopen(fd, "wb") as f:
                 pickle.dump(timeout, f, 1)
-                pickle.dump(value, f, pickle.HIGHEST_PROTOCOL)
+                pickle.dump(value, f, 4)  # 4 was the highest protocol
```
gergelypolonkai: `HIGHEST_PROTOCOL` still exists in Python 3: https://docs.python.org/3/library/pickle.html#pickle.HIGHEST_PROTOCOL

flask_caching/backends/filesystem.py
```diff
@@ -14,14 +14,13 @@
...
 from time import time
-
 from werkzeug.posixemulation import rename
 
 from flask_caching.backends.base import BaseCache
 
 try:
-    import cPickle as pickle
-except ImportError:  # pragma: no cover
+    import _pickle as pickle
```
gergelypolonkai: In Python 3 it is advisable to import `pickle` instead. It will import and use `cPickle` if available.

huangxiaohen2738:
ok. fine

PR #153, opened 2019-12-24
Bugfix: pylibmc expect time as delta time

zoltan-fedor:
Pylibmc expect the timeout to be provided as "number of seconds until the key expire", see http://sendapatch.se/projects/pylibmc/reference.html#pylibmc.Client.set

See more at https://github.com/sh4nks/flask-caching/issues/152

### DIFF
--- a/flask_caching/backends/memcache.py
+++ b/flask_caching/backends/memcache.py
@@ -84,7 +84,19 @@ class MemcachedCache(BaseCache):
     def _normalize_timeout(self, timeout):
         timeout = BaseCache._normalize_timeout(self, timeout)
         if timeout > 0:
-            timeout = int(time()) + timeout
+            # NOTE: pylibmc expect the timeout as delta time up to
+            # 2592000 seconds (30 days)
+            try:
+                import pylibmc
+                using_pylibmc = True
+            except ImportError:
+                using_pylibmc = False
+
+            if not using_pylibmc:
+                timeout = int(time()) + timeout
+            elif timeout > 2592000:
+                timeout = 0
+
         return timeout
 
     def get(self, key):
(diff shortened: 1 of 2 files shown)


flask_caching/backends/memcache.py
```diff
@@ -84,7 +84,19 @@ def _normalize_key(self, key):
     def _normalize_timeout(self, timeout):
         timeout = BaseCache._normalize_timeout(self, timeout)
         if timeout > 0:
-            timeout = int(time()) + timeout
+            # NOTE: pylibmc expect the timeout as delta time up to
+            # 2592000 seconds (30 days)
+            try:
+                import pylibmc
+                using_pylibmc = True
```
gergelypolonkai: `pylibmc` is already imported in `self.import_preferred_memcache_lib`; although `import`, here is a no-op, Iʼd prefer setting `using_pylibmc` there, so itʼs done only once, not for every cache set operation (`_normalize_timeout` is called every time you call `cache.set` is used).

zoltan-fedor: You are quite correct. I have made a change to store the type of memcache library used on an instance attribute, so we can use that to make library-specific adjustments to the code like this one.

zoltan-fedor: It turns out that `self.import_preferred_memcache_lib` is not always called, so I still needed to add a one-time import of `pylibmc` i`n _normalize_timeout` to cover for those scenarios.
Now it minimizes the number of times the `pylibmc` is imported - which is great for performance, but it is not very clear or well structured and definitely not DRY. Maybe it would be time to create a separate method to set `self.mc_library` and instead call that method from these places.

sh4nks:
Thanks for your PR! Sorry that it took me so long to get it merged!

PR #156, opened 2020-01-10
Function source code hashing when generating cache key

komailrk:
Adding source_check option to Cache.cached to have it include the function's source code when forming the hash key to store along with the cached data.

Adding source check functionality to memoize.

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -149,6 +149,8 @@ class Cache(object):
         self.with_jinja2_ext = with_jinja2_ext
         self.config = config
 
+        self.source_check = None
+
         if app is not None:
             self.init_app(app, config)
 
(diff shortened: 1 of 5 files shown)


flask_caching/__init__.py
```diff
@@ -180,6 +182,7 @@ def init_app(self, app, config=None):
         config.setdefault("CACHE_ARGS", [])
         config.setdefault("CACHE_TYPE", "null")
         config.setdefault("CACHE_NO_NULL_WARNING", False)
+        config.setdefault("CACHE_SOURCE_CHECK", False)
```
gergelypolonkai: Please add this new config variable to th edocumentation.

komailrk: Added it to docs/index.rst. Please have a look and let me know if it looks good.

komailrk:
@gergelypolonkai can you have a look at this once more?

PR #173, opened 2020-06-02
Extension for redis cluster mode

AIweimd:
**import redis-py-cluster==2.0.0**
CACHE_TYPE=rediscluster
Add cluster parameter of string type: 
e.g. host1:port1,host2:port2,host3:port3


sh4nks:
Hi,
thanks for your PR - however I don't want to add any more backends if there is not more demand for one because I am the one who has to maintain all those.

I'll leave this PR open for now and if there is more demand I might merge it.

DustinMoriarty:
@sh4nks : I support merging this PR. I just created my own backend for cluster mode to use at my company with Elasticache. I was considering submitting a similar PR.

However, I understand the supporting lots of backends can increase maintenance and testing complexity in an n-combinations sort of way that could get our of hand. In that case, would you prefer that each additional backend be a separate package with more of a plugin/extension architecture emphasis or would it be better for publicly used backends to be maintained within this package?

gergelypolonkai:
Adding more backends is fun and all, but most authors who do so will abandon their addition as long as it’s merged and the maintainers will have to, well, maintain it.

Since this package already supports using external backend classes, i’d probably go on the plugin/separate package way. It’s the best for everyone.

About this PR, it’s not really a new backend, but the extension of an existing one. I’m with it, too.

sh4nks:
I am thinking about having a section with 'user contributed' backends where we do not offer 'official' support and a section with core backends that are supported by us might be sufficient. Especially if the clients/backends are getting more complex like the RedisSentinel one where testing the client took some more time since I had to setup the redis sentinels.

This one looks kinda simple so I guess we can include it :-)

flask_caching/backends/rediscache.py
```diff
@@ -273,3 +273,57 @@ def __init__(
...
+
+        if kwargs.get("decode_responses", None):
+            raise ValueError("decode_responses is not supported by "
+                             "RedisCache.")
+
+        try:
+            from rediscluster import RedisCluster
+        except ImportError:
+            raise RuntimeError("no rediscluster module found")
+        HOST, PORT = 0, 1
```
sh4nks, line 314: while having magic numbers is generally not a good idea - however here it would be more readable when using just node[0].strip() and node[1].strip() in my opinion

AIweimd: Yeah, Magic numbers is not elegant code usually, I like to make it searchable by macro definitions. It’s also good to align the code style here, so I update these code.

sh4nks:
@AIweimd if you could add this I'd merge the PR.


> Please add some documentation about the CACHE_REDIS_CLUSTER setting here: https://github.com/sh4nks/flask-caching/blob/master/docs/index.rst#configuring-flask-caching
> 
> and about the `rediscluster` cache after the `redissentinel` cache here:
> https://github.com/sh4nks/flask-caching/blob/master/docs/index.rst#redissentinelcache

AIweimd:
> @AIweimd if you could add this I'd merge the PR.
> 
> > Please add some documentation about the CACHE_REDIS_CLUSTER setting here: https://github.com/sh4nks/flask-caching/blob/master/docs/index.rst#configuring-flask-caching
> > and about the `rediscluster` cache after the `redissentinel` cache here:
> > https://github.com/sh4nks/flask-caching/blob/master/docs/index.rst#redissentinelcache

AIweimd:
The update document is complete. 
It is an honor to contribute to this project.

docs/index.rst
```diff
@@ -530,82 +544,4 @@ Relevant configuration values
 SpreadSASLMemcachedCache
 ````````````````````````
 
-Set ``CACHE_TYPE`` to ``spreadsaslmemcached`` to use this type.
```
sh4nks, line 533: Please restore this part of the documentation again

AIweimd: Done :-)

PR #176, opened 2020-06-05
uWSGI: Check cache is enabled before use it & fix docstring

johaven:
Check in uWSGI options if cache2 was declared, without this the cache can't be used.


johaven:
Coverage report seems to be is a little demanding but all is ok for me :)

sh4nks:
Thanks!

PR #179, opened 2020-06-22
setup.py: python_requires

singingwolfboy:
Since this package is only compatible with Python 3, we should use the `python_requires` argument in `setup.py` so that tools like `pip` don't try to install it on unsupported versions of Python.

https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

All other changes were automatically made by [`black`](https://github.com/psf/black) in my editor. I can revert them if you prefer.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ class PyTest(TestCommand):
 
     def run_tests(self):
         import shlex
+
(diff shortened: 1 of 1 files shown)


sh4nks:
Thanks for your PR!

PR #194, opened 2020-09-16
less change

huangxiaohen2738:
1.use the `dict` to replace the `OrderedDict` in python3
2.add method `to_str` and `to_bytes`: clearly code

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -16,7 +16,6 @@ import logging
 import string
 import uuid
 import warnings
-from collections import OrderedDict
 
 from flask import current_app, request, url_for
 from werkzeug.utils import import_string
@@ -130,6 +129,20 @@ def function_namespace(f, args=None):
         ins = None
 
     return ns, ins
+
+
+def to_str(str_or_byte):
+    if isinstance(str_or_byte, bytes):
+        return str_or_byte.decode("utf8")
+    return str(str_or_byte)
+
+
+def to_bytes(str_or_byte):
+    if isinstance(str_or_byte, str):
+        return str_or_byte.encode("utf8")
+    elif isinstance(str_or_byte, int):
+        return str(str_or_byte).encode("utf8")
+    return bytes(str_or_byte)
 
 
 def make_template_fragment_key(fragment_name, vary_on=[]):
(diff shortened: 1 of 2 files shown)


flask_caching/__init__.py
```diff
@@ -733,7 +745,7 @@ def _memoize_kwargs_to_args(self, f, *args, **kwargs):
         new_args.extend(args[len(arg_names) :])
         return (
             tuple(new_args),
-            OrderedDict(
+            dict(
```
gergelypolonkai, line 748: Here `OrderedDict` was used to keep the dictionary in, well, ordered. Why do you think ordering is unnecessary here? Also, if it *is* unnecessary, the `sorted` operation of the next line is also not needed (and degrades performance).

flask_caching/__init__.py
```diff
@@ -132,6 +131,20 @@ def function_namespace(f, args=None):
     return ns, ins
 
 
+def to_str(str_or_byte):
+    if isinstance(str_or_byte, bytes):
+        return str_or_byte.decode("utf8")
```
gergelypolonkai, line 136: This can lead to similar problems as the original `str(x)` approach. If `str_or_bytes` contains non-UTF8 byte sequences, you will get an exception.

flask_caching/__init__.py
gergelypolonkai, line 748: `OrderedDict`was used to keep the dictionary, well, ordered. If you think ordering is not needed here, please explain better in the PR description. If it *is* not needed, the `sorted` operation should also be removed in the following line.

gergelypolonkai (changes requested):

flask_caching/__init__.py
huangxiaohen2738: thanks for your reply.
python3.6 later，dict has the order，and used less memory。

flask_caching/__init__.py
huangxiaohen2738: yep. old code has the same prolem。
i think， this function should have the argument of character encoding。

sh4nks:
I do not like this change. There is no guarantee that ``dict`` will stay ordered - this is also an implementation detail. Other Python implementations don't have to follow this.

huangxiaohen2738:
> I do not like this change. There is no guarantee that `dict` will stay ordered - this is also an implementation detail. Other Python implementations don't have to follow this.

https://mail.python.org/pipermail/python-dev/2016-September/146327.html

huangxiaohen2738:
fine. I close this pr.

PR #196, opened 2020-10-03
Issue 189: stop marking wheels as py2 compatible

### LINKED ISSUE #189: The wheel file on pypi.org still advertises itself as py2-compatible
> Filename, size | File type | Python version | Upload date | Hashes
> Flask_Caching-1.9.0-**py2.py3**-none-any.whl (33.9 kB) | Wheel | **py2.py3** | Jun 2, 2020

https://pypi.org/project/Flask-Caching/#files

Support for py2 was dropped in Flask-Caching 1.8.0, so I'm pretty sure the "py2" shouldn't be there in the wheel file name anymore. I'm not sure if the existing wheel files names for 1.8.0 and 1.9.0 can still be updated, but if they can, I'm sure people still maintaining projects in py2 would appreciate it—so pip won't try to install package versions they can't run.

https://www.python.org/dev/peps/pep-0491/#file-format

buckley-w-david:
This PR fixes #189 and no longer marks wheel distributions as universal, as they are no longer py2 compatible.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,9 +6,6 @@ all_files = 1
 [upload_sphinx]
 upload-dir = docs/_build/html
 
-[bdist_wheel]
-universal = 1
-
 [tool:pytest]
 addopts = -vvl --strict --cov=flask_caching --cov-report=term-missing
 norecursedirs = .* _* scripts {args}


gergelypolonkai:
Thanks!

PR #201, opened 2020-11-23
feat: Adds option in memoize to ignore args

Darkheir:
This PR adds an option to ignore some arguments while generating the cache key.

i.e.

```python
@cache.memoize(50, args_to_ignore=["b"])
def big_foo(a, b):
    return a + b + random.randrange(0, 100000)

assert big_foo(5, 3) == big_foo(5, 2)  # True
```


sh4nks:
Looks good. Thanks for your contribution!

Issue #63, opened 2018-03-23
Can you add the CACHE_ configs into Flask config?


Torniojaws:
It's a bit unclear to me if the CACHE_ configs can be put into the regular Flask config, or they are separate from it?

Is this valid?
```
class Config(Object):
    # Unrelated examples
    SECRET_KEY = configfile.get("env", "secret")
    CORS_HEADERS = ["Content-Type", "Authorization", "User"
    # Related to flask-caching
    CACHE_TYPE = "redis"
    CACHE_REDIS_HOST = "127.0.0.1
    # etc


CONFIG = Config()

app = Flask(__name__)
app.config.from_object(CONFIG)
cache = Cache(app)
```

Or should it be separate?
```
class Config(Object):
    # Unrelated examples
    SECRET_KEY = configfile.get("env", "secret")
    CORS_HEADERS = ["Content-Type", "Authorization", "User"


class CacheConfig(Object):
    # Related to flask-caching
    CACHE_TYPE = "redis"
    CACHE_REDIS_HOST = "127.0.0.1


CONFIG = Config()
CACHE_CONFIG = CacheConfig()

app = Flask(__name__)
cache = Cache(app, CACHE_CONFIG)
app.config.from_object(CONFIG)
```

Thanks!

stratosgear:
I will agree with you.  I only got it to work by providing a dictionary to the initializer (like in your second example).

Traditional Flask configuration settings do not seem to work.

sh4nks:
Using your first example should work because of:
```python
base_config = app.config.copy()
if self.config:
    base_config.update(self.config)
if config:
    base_config.update(config)

config = base_config
```

but you are right, the docs should clarify how to configure Flask-Caching.

PR #209, opened 2020-12-31
Extensible serializers support

subnix:
Pickle serializer is not secure and may lead to remote code execution or local privilege escalation. If attacker gains access to cache storage (filesystem, memcached, redis) then he can craft special payload, poison the cache and execute python-code. To avoid the attack we can use a safer serializer (eg. JSON).
I’ve added possibility to use any serializer. All changes are backward compatible and default serializer is pickle.

References:
https://cwe.mitre.org/data/definitions/94.html
https://cwe.mitre.org/data/definitions/502.html
https://davidhamann.de/2020/04/05/exploiting-python-pickle/ (attack example)


oittaa:
Would it make sense to encode bytes in base64 (or something else) with JSON? Having bytes in an object is a relatively common case and JSON can't serialize bytes. If you look at the example below, `cls=BytesEncoder` and `object_hook=as_bytes` could be passed to `json.dumps` and `json.loads` respectively as defaults.


```
import json
from base64 import b64decode, b64encode

BYTES_HINT = "__bytes__"


class BytesEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, (bytes, bytearray)):
            return {BYTES_HINT: True, "bytes": b64encode(obj).decode("ascii")}
        return json.JSONEncoder.default(self, obj)


def as_bytes(obj):
    if BYTES_HINT in obj:
        return b64decode(obj["bytes"])
    else:
        return obj


test_data = dict()
test_data["float"] = 1.5
test_data["bytes"] = b"bytes"
test_data["nested_dict"] = {"1": 1, "2": b"dict_bytes"}
test_data["list"] = [True, "string", b"list_bytes", None]
print(test_data)

jsonstring = json.dumps(test_data, cls=BytesEncoder)
assert test_data == json.loads(jsonstring, object_hook=as_bytes)

```

subnix:
1. It’s a “fake responsibility” that library is responsive for serializing types unsupported by json. We can’t cover all the cases and shouldn’t try.
2. Explicit is better than implicit. This may lead to unexpected behavior by developer and requires strong input validation. For example, application can cache user-provided (or third party) data that may contain special keys (“__bytes__”: True).

All in all, it’s developer’s responsibility, not library’s one.

subnix:
I see 2 options for supporting bytes:
1. Low level. Add some flag (e.g. raw=True) to get/set methods, that allows to bypass serialization.
2. High level. Add decorator (e.g. memoize_bytes) with auto base64 encoding/decoding.

subnix:
My bad, I mean `__bytes__: True`.
```
test_data = (
    {"__bytes__": True},
    {"__bytes__": True, "bytes": "AAA"},
    {"__bytes__": True, "bytes": "!"},
    {"__bytes__": True, "bytes": "Cg=="}
)
for case in test_data:
    jsonstring = json.dumps(case, cls=BytesEncoder)
    assert test_data == json.loads(jsonstring, object_hook=as_bytes)
```

subnix:
@sh4nks are there any updates?

subnix:
I’ve obtained CVE ID (CVE-2021-33026) for this issue.

ndouchin:
@subnix @sh4nks are there any updates on this issue?

We use Flask-Caching and have `safety` in one of our CI job which causes the pipeline to fail.

sh4nks:
Hi, sorry for ignoring this issue for a while. Due to RL I don't have as much time as I'd like for maintenance work. I'll try to take a closer look today or tomorrow.

flask_caching/backends/base.py
```diff
@@ -28,19 +33,47 @@ def iteritems_wrapper(mappingorseq):
...
+
+
 class BaseCache(object):
     """Baseclass for the cache systems.  All the cache systems implement this
     API or a superset of it.
 
     :param default_timeout: The default timeout (in seconds) that is used if
                             no timeout is specified on :meth:`set`. A timeout
                             of 0 indicates that the cache never expires.
+    :param serializer_impl: Pickle-like serialization implementation. It should
```
sh4nks, line 52: I'd much rather have a config variable like ``CACHE_SERIALIZER`` which lets me configure the serializer instead of setting it here

sh4nks (changes requested):

flask_caching/backends/base.py
ndouchin: For this variable `CACHE_SERIALIZER` we would have two values: `pickle` and `json` (and it's up to Flask-Caching internal implementation to reference the proper serializer object)?

Or do you want to have the serializer object directly set in the configuration? But I don't know if it's a good way to handle Flask configuration (we would not be able to load it from a json in this case).

subnix: I will add `CACHE_SERIALIZER` and `CACHE_SERIALIZER_ERROR` config variables with dynamic (string-based) import support, but I think we should keep dependency injection in the base class, otherwise we will need to handle serialization in each implementation.
What about default value? Should we keep pickle as default serializer?

sh4nks: I'd say we stay with 'pickle', issue a warning that this serializer is insecure and starting with 2.0 we switch to the JSON serializer.

subnix:
Pylibmc also uses pickle by default. @sh4nks maybe we should force (de-)serialize data on our own?

https://github.com/lericson/pylibmc/blob/8e783a69cc69fb04b9faad6c61ab43193569ab0f/src/_pylibmcmodule.c#L1245-L1284

potiuk:
Hey @sh4nks - any news about this ? We'v been made aware about he issue at Apache Airflow. It would be great to understand what we can expect here. We might help with some testing etc. if needed.

sh4nks (approved):

ThiefMaster:
**The CVE is rather disruptive and should probably be marked as INVALID** (but good luck doing so, when someone requested a bogus CVE for flask a while ago, it took months until it was fully taken down) as there is **no security vulnerability** here...

Just like your database, the cache should be properly secured and behind authentication (or with proper file system ACLs if file-based). So the only way this is insecure, is when you:

- expose your cache to untrusted users (e.g. unauthenticated redis exposed to the internet);
- have attackers that can write arbitrary files to your disk (and thus poison a file-based cache);
- have a security issue that lets users write arbitrary data to your cache storage backend without going through this library

So when you have a properly secured system and application, it's is a non-issue. Otherwise tools like Celery would need CVEs as well because guess what, they can use Pickle as well.

I'm not even sure how this CVE should be handled by people who are *very* strickt about CVEs ("nothing with open CVE" rules or similar). Because any update containing this change **will not be more secure out of the box**. The default is still pickle, and changing that would require a major version bump and probably break people's applications anyway (some applications are fine with JSON, but I'm sure quite a few do make use of Pickle's ability to serialize almost any Python object). So would it actually be OK to mark the next release as a "fixed version"? I'm not sure.

Also, a CVSS of 9.8 puts this on par with wormable vulnerabilities in widely used network-exposed applications or operating systems. This is simply not true. It doesn't "require no privileges" - you need write access to an internal part of an application. That's probably "requires high privileges".

subnix:
> Just like your database, the cache should be properly secured and behind authentication
> So when you have a properly secured system and application, it's is a non-issue.
Security is not about "should be". It's about people who make mistakes.
What do you think about LPE in Linux? It's not a vulnerability because nobody can gain access to your server?)

> Otherwise tools like Celery would need CVEs as well because guess what, they can use Pickle as well.
Please, check the [documentation](https://docs.celeryproject.org/en/stable/userguide/calling.html#serializers):
- The default serializer is JSON;
- Warning about pickle insecurity;
- Offers cryptography signing;
- By default Kombu will only load JSON messages.
Do you see the difference between Celery and flask-caching? Celery offers native pickle for people who live in an ideal secure world, but for others it offers **choice** and **JSON by default**.

> CVSS of 9.8
I have no idea why it’s so high.

potiuk:
I quite agree 9.8 is wrong. I calculated it and it looks like it should be 9.1:  https://www.first.org/cvss/calculator/3.0#CVSS:3.0/AV:N/AC:L/PR:H/UI:N/S:C/C:H/I:H/A:H  

Is there a way we can appeal that assessment? 

This is a very low risk vulnerability in case of Apache Airflow - one that we can just ignore. In our case we are using "/tmp" as cache storage and if someone can modify it, they can run ANYTHING. This is the same level severity as "have ability to run any program". But wa are pretty special in Apache Airflow where we execute Python code that is managed outside of the main program (by design). So almost by definition we allow to run arbitrary python code provided by our users, so for us this is no increase whatsoever in security risk. I just closed an issue about it in Airflow https://github.com/apache/airflow/issues/16541 - mainly because we are using local storage with "/tmp" folder in Airflow and this configuration is hard-coded. If you have access there, it's a done deal.

However if you use remote cache, it's a bit different story and I agree the issue is critical. And it's rather different from unprotected DB. The problem is that this vulnerability in this case is remotely exploitable. Exploiting this vulnerability, you can MOVE between machines if you gain access to the cache server. You can modify the stored cache and inject your code that then will be executed on the server using it. Unlike the DB access, you can inject CODE which is executed somewhere else - even if the remote system is not hacked. Code that was never supposed to be there. This is not possible by default with any DB I know. This is serious. There should be a way to disable it.

gergelypolonkai (approved):

arihant2math:
> Security is not about "should be". It's about people who make mistakes.

If some got access to the cache storage, they have other things to worry about, especially if its the filesystem.

long76:
any updates?

arpshukl:
@subnix @sh4nks are there any updates on this issue?

flask_caching/serialization.py
```diff
@@ -0,0 +1,34 @@
...
+        if isinstance(result, str):
+            return result.encode()
+        return result
+
+    def dump(self, obj, fp, *args, **kwargs):
+        fp.write(self.dumps(obj, *args, **kwargs))
+
+    def loads(self, obj, *args, **kwargs):
+        if isinstance(obj, bytes):
+            obj = obj.decode()
```
reubano, line 24: what about a `try/except` block here to respect duck typing?

flask_caching/serialization.py
```diff
@@ -0,0 +1,34 @@
...
+
+class BytesSerializer:
+    """Serializer wrapper with auto str2bytes casting"""
+    def __init__(self, serializer):
+        self._serializer = serializer
+
+    def dumps(self, obj, *args, **kwargs):
+        result = self._serializer.dumps(obj, *args, **kwargs)
+        if isinstance(result, str):
+            return result.encode()
```
reubano, line 16: what about a `try/except` block here to respect duck typing?

vedipen:
any update on this? @subnix

dimon222:
Any update?

aduggirala28:
Hi, has anybody found any alternative to this library that doesn't use pickle or doesn't have an open CVE? (We are using flask-caching with plotly/dash, which internally uses the flask-caching object)

davidism:
As stated above, we consider the CVE invalid. You'll want to figure out how to mark those as invalid in whatever scanning tool is reporting them to you. Use of pickle on its own isn't a security issue, it would only be an issue if we were loading arbitrary data, which we're not.

This repo has recently changed maintainers, and the volunteer maintainers are still trying to find time to start working on various issues. There is a PR in cachelib to support serializer customization, and we will be transitioning flask-caching to use cachelib.

fluffy-critter:
Is there any way to get the CVE closed?  I use flask-caching in a library that I provide to others and I'm not thrilled with people being concerned about getting weekly dependabot notifications for this issue which boils down to "if you let people upload code to your server, they can run it."

tvb:
Still not able to close the CVE? My gitlab vulnerability report is also complaining.

fluffy-critter:
For my projects I ended up just telling dependabot to ignore the CVE on the basis of “risk is tolerable to this project.” That doesn’t help downstream users though. :(

long76:
[294](https://github.com/pallets-eco/flask-caching/issues/294)

FLIGHTLAMA:
First of all, thanks for the great work with Flask Cache.
Unfortunately I can't disable warnings in the scanning tool as they are managed centrally at some institutes.
Is there any workaround or fork in the meantime until cachelib is used?

northernSage:
Once #308 is merged, this will need some updated due to [cachelib:#63](https://github.com/pallets/cachelib/pull/63)

subnix:
> Once #308 is merged, this will need some updates due to [cachelib:#63](https://github.com/pallets/cachelib/pull/63)
I see only one way to update this according to changes - close the PR and open a new one :)

@northernSage may be we can implement common serializers (json, pickle with hmac) in `cachelib`? Otherwise every library user will have to implement boilerplate serialization engines.

Note about severity:
According to [Censys](https://search.censys.io/search?resource=hosts&sort=RELEVANCE&per_page=25&virtual_hosts=EXCLUDE&q=services.service_name+%3D+%60REDIS%60+or+services.service_name+%3D+%60MEMCACHED%60), there are 300,000 memched/redis servers exposed to internet, but what about intranet?.. I hope they don't use pickle :). Sure, there isn't a real-time data and honeypots make some part of it.
So we can endlessly discuss pickle (un-)security, but this still holds true - cache storages are not properly secured. The main goal of the CVE is to warn end users about this behaviour. I have decided for me to patch library, because I don't agree with this behaviour, but everyone decides for himself what to do.

rylyade1:
Hi, are there any updates on how this pull request or the corresponding CVE will be handled in the library? Will other serializers (e.g. json) be offered as @subnix suggested?

shahwaizali-10p:
Hi, Any update on when this PR will be merged to fix this CVE issue ?
AWS vulnerability scanner report also complainig.

shahwaizkhan:
Hi, Any update on when this PR will be merged to reslove this CVE issue ?
AWS vulnerability report is also complaining.

northernSage:
@shahwaizkhan, @rylyade1 We are currently going through an internal refactor process, please see #308 mentioned in my previous comment:

> Once #308 is merged, this will need some updates due to [cachelib:#63](https://github.com/pallets/cachelib/pull/63)

Bear in mind that cachelib (which is now flask-caching's backend) already supports custom/configurable serializers, what is left now is exposing this functionality in a convenient way through flask-caching's API, ongoing in https://github.com/pallets-eco/cachelib/issues/136#event-6610243105

bkhakshoor:
Can I get some clarity here, if a user controls the parameters being passed to the memoize function, can they execute code?

aarti9:
Hi all,

I am looking for solution to the vulnerability for flask-caching. Can you please let me know if there is a solution to use? This is a bit urgent so would appreciate any response. Thank you.

Thanks
Aarti

ThiefMaster:
This is not really urgent nor an actual vulnerability. The ONLY case where this can be exploited is if your infrastructure is already compromised and an attacker can write arbitrary data to your cache backend. Please see the comment right above yours as well.

fluffy-critter:
I just submitted an update request to mitre.org with the following description:

> The stated vulnerability is not a vulnerability in flask-caching itself, but only makes flask-caching one intermediate member in a chain of multiple security issues.
> 
> For an attack like this to work, an attacker must:
> 
> 1.  Be able to write arbitrary values to the cache
> 2. Be able to generate a cache key that will collide with a value being read by the application
> 3. Cause the application to read a maliciously-injected value
> 
> Any situation where all 3 of those is true is a situation where the application has larger problems; for example, if someone's able to inject malicious cached rendered pages into a Flask app's cache, then they can make the website say literally anything they want, regardless of whether it involves the execution of remote code. Basically, the Pickle vulnerability follows from a website already being extremely vulnerable (due to conditions 1 and 2 being met).
> 
> There has not been a single confirmed case of this exploit having ever occurred, despite the description being public for two years; the attack is only theoretical, requires a very specific exploit path which requires there to be much worse security issues to be visible in the first place, and it boils down to "if you let people upload code to your server, they can run it."
> 
> This CVE should be rejected, or at least have its severity reduced, with the following base metrics:
> 
> Attack complexity: high (requires multiple failures in security infrastructure)
> User interaction: required (the exploit requires a web request that reads from a poisoned cache item)
> Confidentiality: none (this does not directly result in the disclosure of confidential data)
> Availability: none (it's regarding data in a cache, which is inherently transitory/non-durable)
>

NotoriousPyro:
> Hi, are there any updates on how this pull request or the corresponding CVE will be handled in the library? Will other serializers (e.g. json) be offered as @subnix suggested?

I forked the repo and made a library extending this that replaces the serializers, though its been a while since I worked on it since we went with Next.js instead. I have no idea if it still works. I think it should even be able to serialize bytes to json, by base64 it... again no guarantees and no warranty... https://github.com/NotoriousPyro/flask-caching-json

Legoclones:
Okay I'll be honest I'm surprised no one else has pushed back on the explanation given for why it's not a security vulnerability. I understand and agree it's not a critical severity vulnerability and should not have a 9.8 CVSS score. However, I disagree for the reasons given:

> This CVE should be rejected, or at least have its severity reduced, with the following base metrics:
> 
> Attack complexity: high (requires multiple failures in security infrastructure)
> User interaction: required (the exploit requires a web request that reads from a poisoned cache item)
> Confidentiality: none (this does not directly result in the disclosure of confidential data)
> Availability: none (it's regarding data in a cache, which is inherently transitory/non-durable)

User interaction is not required because no user action outside of the attacker's is required for the vulnerability to be exploited (like in XSS or CSRF vulnerabilities). Confidentiality and availability are both affected because RCE on the underlying platform gives you full access and ability to take down/change the server and see all the data available. 

I go back and forth on attack complexity whether it should be High or Low but I think either one is fine. I think privileges required should be high because you are correct it requires a high level of access to the website. Attack Vector I would also say adjacent because the parts you need to be able to modify to trigger this vulnerability are not remotely accessible. So I would give it a score of [7.6 (High)](https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:A/AC:H/PR:H/UI:N/S:C/C:H/I:H/A:H).

That being said, I think it's totally the maintainer's decision whether it's an accepted risk or whether it should be fixed. It has been made apparent that it won't be changed, and in my opinion that's okay. Personally though, it's still a security vulnerability in my book because *yes* it requires other components to be compromised, but that doesn't prohibit it from being a security risk. It's not as SEVERE of a vulnerability, but it's still a vulnerability. 

Vulnerabilities aren't just something that breaches the outermost layer of an application, and defense in depth is a well-known security principle. Best practice is to secure your application all the way down, including throughout the middle of it. I also have to note that just because a vulnerability hasn't been (knowingly) exploited in the wild doesn't make it any less of a vulnerability. A security professional would account for this when looking at the RISK of the vulnerability, not the SEVERITY of it. This has a high (not critical) severity, but low probability. 

I'm not pushing for any changes, but I feel there is important information missing from this thread for anyone who ends up looking at it.

Issue #188, opened 2020-07-31
Error in how FileSystemCache counts number of files


fnordian:
The set operation in `flask_caching/backends/filesystemcache.py` increases the file count even if the item exists already. This leads to a higher assumed count of files and to more prunes than necessary.

Kl000s:
Is this still open? I would like to give it a try.

sh4nks:
Yeah, I'm happy to review a PR if you want to submit one :-)

Kl000s:
Just submitted the PR :-)

PR #210, opened 2021-01-04
Error in how FileSystemCache counts number of files #188

Kl000s:
The set() method in FileSystemCache now checks if the file is new and only then increases file count. To my understanding, this only happened when caching a None value (this is why in the test case I added None values to the cache). I also fixed two locations where a cache file was deleted without updating the file count (in has() and get() methods).


sh4nks:
Thanks!

PR #218, opened 2021-02-16
Fix add() in RedisCache without a timeout

ThiefMaster:
When there is no timeout, this resulted in the key expiring immediately after setting it, instead of not having an expiry at all.


sh4nks:
Thanks!

Issue #213, opened 2021-02-02
Please make subclassing cache backends easier


ThiefMaster:
I tried using `app.config['CACHE_TYPE'] = 'mypackage.MyRedisCache'` to use a custom subclass of the redis cache backend, and was then surprised that things didn't work. Until I realized that this shouldn't point to the class itself, but the function from `backends/__init__.py`... so if I want to use a custom subclass, I now need to copy&paste about 25 lines of code from there into my application which is of course rather awful.

It would be nice if this factory function was a classmethod on the actual backend class.

sh4nks:
I fully agree! I don't have much time right now as I became a dad recently :-). If you want to submit a PR I'd gladly merge it.

gergelypolonkai:
@ThiefMaster while this gets fleshed out, you can copy the corresponding function from `backends/__init__.py` and set `CACHE_TYPE` to the path of that function.

But indeed, this could work much better, if we can make the `__init__` of the backend classes uniform.

sh4nks:
This is implemented now thanks to @gergelypolonkai. I have published a new version which includes this change. In the future, this will be the only way to use backends.

PR #219, opened 2021-02-17
Move cache factory functionality into the cache backend classes.

gergelypolonkai:
This is a partial solution to #213

### DIFF
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -216,6 +216,11 @@ class Cache(object):
     def _set_cache(self, app: Flask, config) -> None:
         import_me = config["CACHE_TYPE"]
         if "." not in import_me:
+            warnings.warn(
+                "Using the initialization functions in flask_caching.backend "
+                "is deprecated.  Use the a full path to backend classes "
+                "directly.",
+                category=DeprecationWarning)
             import_me = type(self).__module__ + '.backends.' + import_me
 
         cache_factory = import_string(import_me)
--- a/flask_caching/__init__.py
+++ b/flask_caching/__init__.py
@@ -21,6 +21,7 @@ from collections import OrderedDict
 from flask import current_app, request, url_for, Flask
 from werkzeug.utils import import_string
 from datetime import timedelta
+from flask_caching.backends.base import BaseCache
 from flask_caching.backends.simplecache import SimpleCache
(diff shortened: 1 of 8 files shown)


sh4nks (approved):
Looks good to me!

flask_caching/__init__.py
```diff
@@ -215,28 +216,28 @@ def init_app(self, app: Flask, config=None) -> None:
...
-                    "%s is not a valid Flask-Caching backend" % (import_me)
-                )
-        else:
-            cache_obj = import_string(import_me)
+            warnings.warn(
+                "Using the initialization functions in flask_caching.backend "
+                "is deprecated.  Use the a full path to backend classes "
+                "directly.",
+                category=DeprecationWarning)
+            import_me = type(self).__module__ + '.backends.' + import_me
```
ThiefMaster, line 224: `type(self).__module__` is breaking any application that subclasses this class since it will then resolve to the module where the subclass is. So I think this should remain as a hardcoded `'flask_caching.backends.' + import_me`

gergelypolonkai: This line only runs if the cache backend you specify in your config (in `CACHE_TYPE`) doesn’t contain a dot (see line 218). If you define your own backend, it will be of the form `your_package.your_module.YourClass`, so `import_me` won’t be mangled for you.

ThiefMaster: i was more thinking about the built-in cache types. `CACHE_TYPE = 'null'` for example. Why would I ever want that to resolve to something different than the built-in one?

gergelypolonkai: In that case, `null` gets expanded to `flask_caching.backends.null`; `self.module` is always `flask_caching` in this context; itʼs written this way so even if we rename the whole module, we donʼt have to modify it here.

ThiefMaster: It isn't if you subclass the `Caching` class in your own module. Then it resolves to that one :/

gergelypolonkai: Ah, so thatʼs what you meant! Sorry for not getting it earlier.

Yes, you are right. I didnʼt consider that when writing the code because i couldnʼt imagine a use case for that. Unless you want to submit a fix, iʼll look into it later this week.

PR #225, opened 2021-03-01
Run tests under uWSGI

gergelypolonkai:


gergelypolonkai:
Since we move `uwsgicache` to `contrib`, this PR is no longer necessary.

PR #229, opened 2021-03-02
Return first existing cache instance

mo7amed-3bdalla7:
Return first existing cache instance in case if required one doesn't exist


sh4nks:
If you have a different caching instance you should rather use the other one directly instead of working around it.

PR #235, opened 2021-03-05
use the cache factory class to register or get cache-function-loader

huangxiaohen2738:
https://github.com/sh4nks/flask-caching/issues/234


sh4nks:
Thanks for your contribution, however this classmethod is just here to transition to the class based definition of caching backends and will be removed for 2.0.

PR #236, opened 2021-03-06
Add test_generic_get_bytes

oittaa:
Some serializers, like JSON, don't support byte strings and those need to be handled as a special case. It might be useful to test for that anyway.


sh4nks:
Thanks!

Issue #243, opened 2021-04-01
Cannot delete cached item if query_string = True


gaw89:
Currently if the cached decorator is used with `query_string=True`, you cannot clear the cached value manually using `cache.delete` by simply providing the `key_prefix` and URL for the cached route.  This is because when `query_string` is set, the `cached` decorator uses `_make_cache_key_query_string` to create the key, as indicated [here](https://github.com/sh4nks/flask-caching/blob/5227fc50b31c936ad530c9dbb0d94c08cb6c9eab/flask_caching/__init__.py#L391).

**This is not indicated anywhere in the docs** (that I could find) **and makes for a very painful user experience.**  I spent several hours trying to figure out why my cache was not clearing when I told it to.  [Here is a gist](https://gist.github.com/gaw89/e7028ddab02e546e9b214b750ab1f216) demonstrating the behavior.

I have several ideas to remedy this, and I am happy to submit a PR.  However, I would like to solicit feedback on potential solutions prior to spending time working on one to maximize the chances of having it incorporated.  Here are my ideas in (what I think is) best to worst order:

1. Add `query_string` keyword to `cache.delete` which would instruct the delete method to append the hashed query parameters to the provided key.  (*This seems to me to be the most complimentary to the current behavior when using the `query_string` parameter*)
2. Add some kind of `regex_delete` method in addition to the standard delete method that will remove all keys that contain a provided substring.
3. Make the docs **abundantly** clear that, if `query_string` is set to `True`, the `cache.delete` method will not work and the user must figure out how to clear that item.
4. Re-write the `cache.delete` method to search for both the provided key and the provided key plus the hashed query parameters. (**this seems like a bad idea to me that would have unintended consequences**)

I look forward to seeing feedback from the group.

sh4nks:
Hi, not sure if you are still interested in this being fixed/discussed. The only way I can see how this might work is to provide the query string to a delete method. So similar to delete_memoized something like delete_cached. I am a bit hesitant to grow the public facing API even more as the code already is... not nice haha

I believe something like cache.delete(my_cached_method.make_cache_key()) might work

PR #244, opened 2021-04-01
Allow use of query_string in cache.delete

gaw89:
This is a proposed implementation of the first option I presented in #243.

This allows you to provide the `query_string` keyword to `cache.delete` which instructs the delete method to append the hashed query params to the provided key.  This allows for successful deletion of an item that was cached with `query_string=True` provided the request query params are the same as those used to create the cached value initially. 

I suppose that caveat will need to be called out specifically in the docs.  For my purposes, this solution works.  I am certainly open to any suggestions/changes that the maintainers think would be advisable.

Thanks for considering!


gaw89:
Thinking through this a bit more, this is a bad solution.

Reworking it now...

PR #254, opened 2021-05-26
Apache Ignite backend support

sdarlington:
Currently, draft as it needs the development/next version of the Python thin-client. But comments welcomed.

### DIFF
--- a/flask_caching/backends/__init__.py
+++ b/flask_caching/backends/__init__.py
@@ -25,6 +25,7 @@ from flask_caching.backends.rediscache import (
 )
 from flask_caching.backends.simplecache import SimpleCache
 from flask_caching.backends.uwsgicache import UWSGICache
+from flask_caching.contrib.ignitecache import IgniteCache
 
 
 __all__ = (
@@ -39,6 +40,7 @@ __all__ = (
     "gaememcached",
     "saslmemcached",
     "spreadsaslmemcached",
+    "ignite",
 )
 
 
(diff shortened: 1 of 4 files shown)


flask_caching/backends/__init__.py
```diff
@@ -25,6 +25,7 @@
 )
 from flask_caching.backends.simplecache import SimpleCache
 from flask_caching.backends.uwsgicache import UWSGICache
+from flask_caching.contrib.ignitecache import IgniteCache
```
sh4nks, line 28: Don't include your backend here. This is only for officially supported backends and not for user contributed backends.

flask_caching/contrib/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .ignitecache import IgniteCache
+
+__all__ = ['IgniteCache']
```
sh4nks, line 3: I'd like to only use the full important path for contributed caches -- without importing the cache class here.

tests/test_init.py
```diff
@@ -11,6 +11,7 @@
     SASLMemcachedCache,
     SimpleCache,
     SpreadSASLMemcachedCache,
+    IgniteCache,
```
sh4nks, line 14: We don't provide tests in the official testsuite for use contributed backends.

sh4nks (changes requested):
Hi, I left a few comments on your draft if you want to have it included in Flask-Caching.

PR #260, opened 2021-07-12
Updated simple to SimpleCache in docs

shoaib30:
changed simple to SimpleCache since simple is deprecated


sh4nks:
nice catch, thanks!

PR #261, opened 2021-07-17
[pickle] Dump using the highest available protocol

john-bodley:
This PR ensures that all backends use the same protocol (the highest available) for pickling the payload to cache. Some backends currently use the default (4 for Python 3.8) whereas some use 2 and others use the highest protocol available (maximum of 5). 

Per the `pickle` [documentation](https://docs.python.org/3/library/pickle.html#data-stream-format),

> Protocol version 5 was added in Python 3.8. It adds support for out-of-band data and speedup for in-band data. Refer to PEP 574 for information about improvements brought by protocol 5.

it seems there is merit in utilizing the highest available protocol for consistency and increased performance.

Note that it seems the protocol is encoded within the dumped object, i.e., one does not need to re-specify the protocol when loading via `pickle.loads(...)`, thus this should be a non-breaking change.

### DIFF
--- a/flask_caching/backends/memcache.py
+++ b/flask_caching/backends/memcache.py
@@ -327,7 +327,7 @@ class SpreadSASLMemcachedCache(SASLMemcachedCache):
         # I didn't found a good way to avoid pickling/unpickling if
         # key is smaller than chunksize, because in case or <werkzeug.requests>
         # getting the length consume the data iterator.
-        serialized = pickle.dumps(value, 2)
+        serialized = pickle.dumps(value, pickle.HIGHEST_PROTOCOL)
         values = {}
         len_ser = len(serialized)
         chks = range(0, len_ser, self.chunksize)
--- a/flask_caching/contrib/googlecloudstoragecache.py
+++ b/flask_caching/contrib/googlecloudstoragecache.py
@@ -102,7 +102,7 @@ class GoogleCloudStorageCache(BaseCache):
         full_key = self.key_prefix + key
         content_type = "application/json"
         try:
-            value = json.dumps(value)
+            value = json.dumps(value, pickle.HIGHEST_PROTOCOL)
         except (UnicodeDecodeError, TypeError):
             content_type = "application/octet-stream"
         blob = self.bucket.blob(full_key)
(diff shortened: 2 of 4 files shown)


flask_caching/backends/memcache.py
```diff
@@ -327,7 +327,7 @@ def _set(self, key, value, timeout=None):
         # I didn't found a good way to avoid pickling/unpickling if
         # key is smaller than chunksize, because in case or <werkzeug.requests>
         # getting the length consume the data iterator.
-        serialized = pickle.dumps(value, 2)
```
john-bodley, line 330: I'm uncertain why this was hardcoded as `2`.

sh4nks:
While I am generally for this PR I am not sure if this might break some applications?

flask_caching/contrib/googlecloudstoragecache.py
```diff
@@ -102,7 +102,7 @@ def set(self, key, value, timeout=None):
         full_key = self.key_prefix + key
         content_type = "application/json"
         try:
-            value = json.dumps(value)
+            value = json.dumps(value, pickle.HIGHEST_PROTOCOL)
```
sh4nks, line 105: pickle isn't used here

sh4nks (changes requested):

flask_caching/contrib/googlecloudstoragecache.py
john-bodley: Apologies. I got a tad carried away.

john-bodley:
@sh4nks regarding,

> While I am generally for this PR I am not sure if this might break some applications?

per [here](https://realpython.com/python-pickle-module/):

> Note: Newer versions of the protocol offer more features and improvements but are limited to higher versions of the interpreter. Be sure to consider this when choosing which protocol to use.
> 
> To identify the highest protocol that your interpreter supports, you can check the value of the pickle.HIGHEST_PROTOCOL attribute.

My take is that by using `pickle.HIGHEST_PROTOCOL` (as opposed to an explicit version—other than -1) ensures safely on a per interpreter level.

john0isaac:
we use the highest protocol in dumping now in CacheLib. this pull request was created prior to using CacheLib as the backend provider.
I think it's no longer valid since there is no serializers here in flask-caching + it's already done in CacheLib.
@sh4nks it can be closed.

Issue #129, opened 2019-07-17
flask with MethodView 


Mr-liu-yang:
How to do cache using the pluggable view of flask ？

Mr-liu-yang:
Anyone else can help me?

Issue #57, opened 2018-01-25
Use with flask.views.View


nicoulaj:
In my application views are implemented by subclassing View, eg:
```
class SomeView(View):
    def dispatch_request(self, *args, **kwargs):
        # impl
```

Adding the annotation on `dispatch_request` does not work, it would be nice if it could be supported.

sh4nks:
I'm happy to review a PR if you want to submit one.

PR #267, opened 2021-08-08
add test and docs for caching pluggable view (flask.views.View) classes

snarfed:
Originally requested in #57, #129. Looks like this is already supported, by decorating the `dispatch_request()` method. Is that expected? Assuming so, here's a brief mention in the docs and a test. Thanks in advance!


northernSage (approved):
Thanks for your contribution!

PR #286, opened 2021-10-20
Support generator in func, but view

ponytailer:
Only support generator in func  with cached/memoized,  but not support views

- [x] Support generator in func, but view
- [x] Fix lint
- [x] Create module called utils,  inspect function 

Checklist:


northernSage (approved):
@ponytailer LGTM, thanks for the contribution!

Issue #295, opened 2021-11-28
Support Dynamic Cache Timeouts Based on Response


phil-lopreiato:
I'm looking for a way to dynamically configure the cache timeout for a view, for the case where the desired timeout depends on the request parameters.

Imagine a page where the data within is normally very static (and can therefore be cached for days at at time), but with a short window of time where it is expected to update very frequently and thus requires a much shorter expiration.

Currently, this pattern is impossible to express with this library

PR #296, opened 2021-11-28
Support dynamic cache timeouts from flask responses

phil-lopreiato:
This diff adds custom `Response` type inheriting from `flask.Response` but adds a `timeout` property. The `cached` decorator will check if the response type is `CachedResponse` and if so, and if the timeout is nonnull, that value will take precedence over the static value stored on the decorated function.

This allows the timeout to be determined dynamically within the view, for example, based on the input args.

Closes https://github.com/pallets-eco/flask-caching/issues/295


northernSage:
Thanks for the contribution @phil-lopreiato, could you please add an entry to `CHANGES.rst` describing the new feature?

northernSage (approved):

PR #356, opened 2022-05-29
Added proxy to directly check if cache `has` a key

### LINKED ISSUE #355: Add support for `has` proxy
There are several proxy functions that allow commands on the underlying cache object. Examples are:
```python3
cache.set(*args, **kwargs)
cache.get(*args, **kwargs)
cache.delete(*args, **kwargs)
```
I'm just adding support for `cache.has(key)` that checks if the `cache` has an existing `key`.

olumidesan:
Added support for `has` proxy

- fixes #355 

Checklist:

- [x] Add `.. versionchanged::` entries in any relevant code docs.


northernSage (approved):
Thanks for the contribution!

PR #365, opened 2022-06-09
Remove overrides of methods inherited from `cachelib.FileSystemCache`

### LINKED ISSUE #363: FileSystemCache broken in 1.11.1
FileSystemCache appears to be broken in 1.11.1 (pickle errors) as a result of the `cachelib` integration.

To reproduce the bug, just use the `FileSystemCache` as a backend.

When attempting to read an existing key from the cache, you will get an error every time:

```
Traceback (most recent call last):
  File "test.py", line 30, in get
    if (c := cache.get(key)) is None:
  File "/usr/local/lib/python3.9/site-packages/flask_caching/__init__.py", line 190, in get
    return self.cache.get(*args, **kwargs)
  File \"/usr/local/lib/python3.9/site-packages/flask_caching/backends/filesystemcache.py\", line 120, in get
    pickle_time = pickle.load(f)
ValueError: unsupported pickle protocol: 198
``` 

I believe the bug is caused by the following:

`FileSystemCache.get()` is overloaded (inherited from `CachelibFileSystemCache`) but doesn't replicate the same pickle logic as the inherited and non-overloaded `CachelibFileSystemCache.set()` method. In particular, the `set()` method stores the `pickle_time` in the first 4 bytes and then the actual pickle data. The overloaded `get()` method simply tries to use `pickle.load()` on the whole thing and blows up due to the first 4 bytes not being "pickle" data.

Environment:

- Python version: 3.10.4
- Flask-Caching version: 1.11.1

gumptionthomas:
Improves compatibility with `cachelib.FileSystemCache` by removing methods and tests that depend on internal implementation details.

- fixes #363

Checklist:

- [X] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [X] Add or update relevant docs, in the docs folder and in code.
- [X] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.
- [X] Add `.. versionchanged::` entries in any relevant code docs.
- [X] Run `pre-commit` hooks and fix any issues.
- [X] Run `pytest` and `tox`, no tests failed.


gumptionthomas:
"One of my most productive days was throwing away 1000 lines of code."

A PR that only removes code (other than `CHANGES`) and fixes tests in the process is a rarity :-)

sorenwacker:
Hi, I am still having issues with the file system cache. In fact I had to downgrade to 1.10.1.

northernSage:
Thanks a lot for the contribution @gumptionthomas, it's greatly appreciated. I'll be reopening this with a more descriptive branch name, though. Also, since implementation details are being removed, this is a breaking change for users who rely on said behavior so we need a major bump in `CHANGELOG.md`.

Issue #362, opened 2022-06-02
Redis Sentinel backend not working after Cachelib rewrite (1.11.0)


peter-slovak:
I'm having problems using `CACHE_TYPE="redissentinel"` since `Flask-Caching` switched to `cachelib`. Instead of querying any of the read/write clients, the `get_many()` method is trying to contact the default `localhost:6379` address. This behavior appeared strictly after upgrading from `1.10.1` to `1.11.1` and disappeared after downgrade.

Related settings in Flask configuration:
```python
    CACHE_TYPE = "redissentinel"
    CACHE_DEFAULT_TIMEOUT = 300
    CACHE_KEY_PREFIX = "cache_"
    CACHE_REDIS_SENTINELS = [("tasks.redis-sentinel", 26379)]
    CACHE_REDIS_SENTINEL_MASTER = "mymaster"
```

From the brief research I did, this is what looks suspicious to me: https://github.com/pallets-eco/flask-caching/pull/308/files#diff-d15a59e5c66734a6b6d638efe7582b3a147d5a4213dc3b3ca4e3672c50607ae1R62

You see, in `1.10.1`, the `get_many()` method queried `self._read_clients` that was initialized in `RedisCache.__init__()` and overloaded in `RedisSentinelCache.__init__()`. However in `1.11.1` the `get_many()` method is now inherited from the `CachelibRedisCache` superclass, which internally uses `self._client` initialized to default values (`localhost:6379`). Although we keep overloading `self._write_client` and `self._read_clients` in `RedisSentinelCache.__init__()`, this has no effect.

Environment:

- Python version: 3.10
- Flask-Caching version: 1.11.1

peter-slovak:
FWIW, the situation is probably the same for `RedisClusterCache`, although I haven't tested.

lexisstv:
Confirm - Sentinel stop working. 
May be problems here:
RedisSentinelCache.__init__() -> RedisCache.__init__() -> (ADDED in 1.11.1 from cachelib import RedisCache as CachelibRedisCache) CachelibRedisCache.__init__() -> ... Redis client initializes connection pool with default arguments (localhost)

northernSage:
Already writing a fix fir this. It may take a few days due to lack of time, but it will be out soon. Meanwhile, please downgrade as suggested.

peter-slovak:
Thank you for the effort to fix this @northernSage ! :hugs: :+1:

PR #372, opened 2022-06-26
Fix redis sentinel backend interface with cachelib

northernSage:
Follow up for of https://github.com/pallets-eco/cachelib/pull/159

This makes it so `flask_caching.backends.RedisSentinelCache` can `flask_caching.backends.RedisClusterCache` will now override `cachelib.RedisCache._write_client` and `cachelib.RedisCache._read_client `, addressing the problem described in #362

- [x] override read/write clients from `RedisSentinelCache`
- [x] add changelog
- [x] update package requirements in `setup.py` and requriements dir
- [x] add tests covering bug for RedisSentinelCache


quito-inv:
Hi, was this fix released? I am not seeing it in the changelog and I'm getting this error that looks like is fixed by this PR
```
.venv/lib/python3.8/site-packages/flask_caching/backends/rediscache.py", line 62, in __init__
    self._write_client = self._read_clients = self._client
AttributeError: 'RedisCache' object has no attribute '_client'
```
This error happened after upgrading to v1.11.1 from v1.11.0. 

I think this PR fixes it but is it possible that it wasn't released?

northernSage:
@quito-inv this has been release in [2.0.0](https://github.com/pallets-eco/flask-caching/blob/master/CHANGES.rst#version-200) along with other improvements/fixes.

quito-inv:
@northernSage oh great, but that means that the latest v1.x are broken? Neither work, at least for me. In 1.11.0 I get an error because cachelib doesn't exist and in 1.11.1 I get the error I mentioned before

Issue #384, opened 2022-07-18
Type errors in the code


rnestler:
When type checking flask-caching code with mypy it shows errors:
```
src/flask_caching/contrib/uwsgicache.py:40: error: Cannot find implementation or library stub for module named "uwsgi"
src/flask_caching/contrib/googlecloudstoragecache.py:12: error: Cannot find implementation or library stub for module named "google.auth.credentials"
src/flask_caching/contrib/googlecloudstoragecache.py:13: error: Cannot find implementation or library stub for module named "google.cloud"
src/flask_caching/backends/rediscache.py:233: error: Cannot find implementation or library stub for module named "rediscluster"
src/flask_caching/backends/memcache.py:105: error: Skipping analyzing "pylibmc": module is installed, but missing library stubs or py.typed marker
src/flask_caching/backends/memcache.py:105: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
src/flask_caching/__init__.py:198: error: Incompatible return value type (got "Optional[bool]", expected "bool")
src/flask_caching/__init__.py:214: error: No return value expected
```

 * Set up flask developer environment as explained in `CONTRIBUTING.rst`
 * Install mypy: `pip install mypy`
 * Run mypy: `mypy src`

Since the package advertises type annotations with the `py.typed` file, I expect the type annotations to be correct. Ideally it would even pass the `mypy --strict` check, but this currently throws 190 errors.

Environment:

- Python version: 3.10.5
- Flask-Caching version: Current master at 6d3e3a34fa6a3edf55f924ab32c98c64597aaae6

PR #385, opened 2022-07-18
Fix type annotation errors in Cache

rnestler:
Fix type annotations in the `Cache` class as pointed out by mypy.

- See #384 (this PR alone does not fix it completely)

Checklist:


northernSage (approved):
Thanks for the contribution! Also, that makes me thing we should probably add a type hint check step to our GH workflow in order to avoid this type of issue in the future (that's for another time, though :bowtie:)

PR #401, opened 2022-09-27
Callable timeout

thejeremyjohn:
Changes:

- added support for callable `timeout` in `@cached` and `@memoized`
- modified try-except import strategy re pytest-xprocess

A callable `timeout` serves a similar purpose to `CachedResponse` (introduced in [PR #296](https://github.com/pallets-eco/flask-caching/pull/296)) but does not replace that functionality entirely. These are the differences:

| callable timeout     | CachedResponse                                         |
|----------------------|--------------------------------------------------------|
| @cached OR @memoized | @cached                                                |
| Any rv type          | Response-like rv type                                  |
| only knows the decorated function's output | knows all locals within the decorated function |

A `CachedResponse` always takes priority over callable `timeout` if it is the sole rv type so there's no point in using both in that case. However if the rv type is conditional they could be complimentary.

NOTE: Prior to this PR, `tests/test_backend_cache.py::TestRedisCache::test_generic_inc_dec` are skipped if pytest-xprocess can't be imported but with that import fixed (1st commit in this PR), they actually fail. (I did not investigate further. FWIW I'm running `Redis version=7.0.0`) That is why the below box is unchecked, but all other tests pass.

Checklist:

- [x] Add `.. versionchanged::` entries in any relevant code docs.


northernSage:
Hey @thejeremyjohn, thanks for the contribution! It looks like we are failing some tests, could you please take a look?

thejeremyjohn:
Hi @northernSage. Yw -- I'm glad to do it! Yeah, that's consistent with what I noted in the PR. Whatever is failing there is not related to the callable timeout changes. 

It's only tripping because those tests are running rather than being skipped due to `pytest-xprocess` not importing, which I addressed in the first commit in this PR. I _could_ undo that such that those tests are again skipped; that's no problem for me since it's not connected to the callable timeout enhancement.

thejeremyjohn:
@northernSage Please let me know if that ^^^ is the preferred way to go.

sh4nks:
I merged your PR by cherry-picking your commits here: #669

PR #431, opened 2022-12-26
bug fix: make cached view function make_cache_key attr writeable

### LINKED ISSUE #97: make_cache_key in @cached is not working
The flask-caching module is:

```
$ pip list | grep Cach
Flask-Caching    1.6.0
```

And the test code is:

```
from api import cache

def zzz(*args, **kwargs):
    return 'aaa'

@cache.cached(timeout=3600)
def aaa():
    a = 1
    if a:
        return 222
    else:
        return 333

aaa.make_cache_key = zzz
```

snarfed:
fixes #97. this matches how memoize already does the same thing:

https://github.com/pallets-eco/flask-caching/blob/b526e03edf0602d10d9cf938b955a59b521636ad/src/flask_caching/__init__.py#L838

Checklist:

- N/A Add or update relevant docs, in the docs folder and in code.
- N/A Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Version 2.0.2
 -------------
 
 - migrate ``flask_caching.backends.RedisCluster`` dependency from redis-py-cluster to redis-py
+- bug fix: make the ``make_cache_key`` attributed of decorated view functions writeable. :pr:`431`, issue `#97`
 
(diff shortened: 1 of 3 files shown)


northernSage (approved):
LGTM, thanks @snarfed

Issue #307, opened 2022-01-03
cachelib as backend


northernSage:
`cachelib` has seen some development since  `flask-caching` used to have it as its backend, including bug fixes, optimization, typing and new features which could be interesting to have in `flask-caching`. Since I'm currently working on re-integrating both projects, I have opened this issue in order to document some key points and also open it for discussion so others have some insight on how the next release is shaping out.

northernSage:
To start with, `flask-caching.backends.FileSystemCache` allows users to specify a `hash_method` used for hashing keys (proposed in #65 and implemented in #77). This seems like an interesting feature to have, so I have already written and merged a PR (https://github.com/pallets/cachelib/pull/107), adding this to `cachelib` (to be released in [0.6.0](https://github.com/pallets/cachelib/blob/main/CHANGES.rst#version-060)).

northernSage:
From #64, it seems early implementations of `cachelib` used to _short-circuit_ removal of multiple keys on the first failed attempt when using `delete_many` for `SimpleCache` and `FileSystemCache`. Since users could potentially be relying on this unspecified behavior, the `ignore_errors` flag was added to flask-caching in #94. The flag defaults to False, allowing the user to opt out of the short-circuiting behavior by setting `CACHE_IGNORE_ERRORS` to `True`.

`cachelib` no longer behaves like this and will always attempt removal of all keys passed into `delete_many` by default, so I'm considering dropping `CACHE_IGNORE_ERRORS` and related functionality once the integration is done. This will probably be a major release (2.0.0).

PR #456, opened 2023-02-11
Use delete_many from cachelib. Remove CACHE_IGNORE_ERRORS flag.

EugeneTorap:
We want to upgrade `flask-caching` lib from 1.10.1 to 2.0.2 in apache/superset. PR is https://github.com/apache/superset/pull/22925

But we have a problem with new `delete_many` logic for `RedisCache` in `BaseCache`.
Old `delete_many` logic in 1.10.1 version:
```python
if self.ignore_errors:
    return all([self.delete(key) for key in keys])
return all(self.delete(key) for key in keys)
```

New `delete_many` logic in 2.0.2 version:
```python
deleted_keys = []
for key in keys:
    if self.delete(key):
        deleted_keys.append(key)
    else:
        if not self.ignore_errors:
            break
return deleted_keys
```

In the first case, all cache keys will be deleted, even if some cache keys do not exist.
In the second case, if we have a non-existent first cache key, we will not delete all existing cached keys.

- @northernSage As you mentioned in #307:

> cachelib no longer behaves like this and will always attempt removal of all keys passed into delete_many by default, so I'm considering dropping CACHE_IGNORE_ERRORS and related functionality once the integration is done. This will probably be a major release (2.0.0).

I've removed CACHE_IGNORE_ERRORS flag and now we use `delete_many` from `cachelib`.

Checklist:

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -375,12 +375,6 @@ The following configuration values exist for Flask-Caching:
 ``CACHE_DEFAULT_TIMEOUT``       The default timeout that is used if no
                                 timeout is specified. Unit of time is
                                 seconds.
-``CACHE_IGNORE_ERRORS``         If set to any errors that occurred during the
-                                deletion process will be ignored. However, if
-                                it is set to ``False`` it will stop on the
-                                first error. This option is only relevant for
-                                the backends **filesystem** and **simple**.
-                                Defaults to ``False``.
 ``CACHE_THRESHOLD``             The maximum number of items the cache
                                 will store before it starts deleting
                                 some. Used only for SimpleCache and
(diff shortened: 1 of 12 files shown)


EugeneTorap:
@northernSage ???

EugeneTorap:
@ashsharma96 @northernSage Any news?

EugeneTorap:
@sh4nks @northernSage Can you help me to fix the problem?

EugeneTorap:
@davidism @sh4nks @northernSage ^^

EugeneTorap:
@davidism @sh4nks @northernSage ^^

EugeneTorap:
@davidism @sh4nks @northernSage ^^

PR #477, opened 2023-04-18
Make cachelib requirement consistent to 0.10.2

hnformentin:
I think the checklist does not apply for such a simple PR.
The requirements for cachelib does not match considering [dev requirements](https://github.com/pallets-eco/flask-caching/blob/d7037dfbaccee828f000ff246dbb3421511f6542/requirements/dev.txt#L19) and setup requirements, modified in this PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,5 +3,5 @@ from setuptools import setup
 # Metadata goes in setup.cfg. These are here for GitHub's dependency graph.
 setup(
     name="Flask-Caching",
-    install_requires=["cachelib >= 0.9.0, < 0.10.0", "Flask < 3"],
+    install_requires=["cachelib == 0.10.2", "Flask < 3"],
 )


hnformentin:
@northernSage could you please review this one?

northernSage:
Hey, thanks for taking the time to open the PR, but a bit more work will be needed in order to make flask-caching 100% compatible with latest `cachelib` since, for example,  it supports `dynamoDB` for which flask-caching still does not offer an interface (see https://github.com/pallets-eco/flask-caching/pull/441). In other words, we would need to implement such interface here and also extend our test suite to cover that new backend. Without that, I don't see much of a reason to bump `cachelib` given that can be even confusing for users who might consider `flask-caching` for use with dynamo only to find that it's not supported even with a version of cachelib that supports it.

dchevell:
Same issue - I don't necessarily agree that just because an underlying dependency supports new things, this library can't ship that dependency without supporting them too. I'd rather avoid situations where some dependency is stuck on an old version, even if all the tests pass on the latest one

northernSage:
closed in favour of #568

PR #497, opened 2023-06-22
Update type annotations for cache proxy class

### LINKED ISSUE #467: Return type annotation for `get()` is too restrictive
Attempting to call `cache.get()` on anything that does not return a string results in a type error when running mypy:

```py
class CustomClass:
    pass

def get_cached_class() -> Optional[CustomClass]:
    return cache.get("key")

$ mypy test.py
> error: Incompatible return value type (got "Optional[str]", expected "Optional[CustomClass]")  [return-value]
```

If we take a look at the type in cachelib (to which this method simply proxies), it actually returns `typing.Any` (which makes sense, because the implementation handles serialization/deserialization using pickle):

https://github.com/pallets-eco/cachelib/blob/c00d97f757364ad99e1fd31b14f5f8f5aacd714c/src/cachelib/base.py#L21

A simple fix would be to just swap out the return type for `Any`.

Environment:

- Python version: 3.8
- Flask-Caching version: 2.0.2

tonybaloney:
Fixes #467 

Also noticed some other methods are incorrectly typed.

### DIFF
--- a/src/flask_caching/__init__.py
+++ b/src/flask_caching/__init__.py
@@ -15,18 +15,13 @@ import logging
 import uuid
 import warnings
 from collections import OrderedDict
-from typing import Any
-from typing import Callable
-from typing import Optional
-from typing import Tuple
-from typing import Union
+from typing import Any, Dict, List, Callable, Optional, Tuple, Union
 
 from flask import current_app
 from flask import Flask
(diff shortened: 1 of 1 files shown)


northernSage:
Thanks for contributing :bowtie:

PR #519, opened 2023-08-20
make_cache_key docs and example added

### LINKED ISSUE #222: New argument "make_cache_key" missing in the docs
Since version 1.9 this argument is supported but not mentioned anywhere in the docs (except change log).

I can add it if I find the time. Though I am unsure where it is available, just using view caching for now:

I'd add something like that:

```
def make_key():
    user_data = request.get_json()

    return ",".join([f"{key}={value}" for key, value in user_data.items()])


@app.route("/hello", methods=["POST"])
@cache.cached(timeout=60 * 60, make_cache_key=make_key)
def some_func():
    ....

```

+ some text around it

ninadpchaudhari:
`make_cache_key` was missing from the docs.
adding for reference. 
Pl see

https://github.com/pallets-eco/flask-caching/issues/222

- fixes #222 

Checklist:

- [N/A] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [N/A] Add `.. versionchanged::` entries in any relevant code docs.
- [N/A] Run `pre-commit` hooks and fix any issues.
- [N/A] Run `pytest` and `tox`, no tests failed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 Changelog
 =========
 
+- Added docs and example for make_cache_key
+
 Version 2.0.2
 -------------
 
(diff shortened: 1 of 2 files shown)


northernSage (approved):
Thanks 👍🏻

PR #547, opened 2023-10-18
Update cache type names in index.rst

tess-lu:
Replace deprecated cache type names `filesystem` and `simple` with new names `FileSystemCache` and `SimpleCache`.


northernSage (approved):
thanks for helping out @tess-lu !

PR #558, opened 2024-04-11
Respect kwargs when args_to_ignore is set in memoize

### LINKED ISSUE #557: Respect `args_to_ignore` for keyword arguments when using `memoize`
I'm not sure if this is a bug or the intended behavior, but it came as a surprise to me. The `args_to_ignore` is not respected when the ignored argument is passed with a keyword. The following code snippet illustrates the problem:
```python
from flask import Flask

simple_cache = Cache(app=Flask("Test"), config={"CACHE_TYPE": "SimpleCache"})


@simple_cache.memoize(args_to_ignore=["irrelevant"])
def square(n: int, irrelevant: int) -> int:
    print(f"Cache miss for {n}-{irrelevant}")
    return n * n


square(5, 10)
square(5, 11)
square(5, irrelevant=10)
```
The output of this snippet is
```
Cache miss for 5-10
Cache miss for 5-10
```
In my opinion, the expected behavior is that only the first call results in a cache miss. In case this is the intended behavior, I would treat this issue as a feature request to add a way to ignore arguments that are passed with a keyword.

Practical context: I have a Flask app that relies on SQLAlchemy and has functions where a `Session` object is passed as a kwarg and I want it ignored.

Environment:

- Python version: `3.9.4`
- Flask-Caching version: `2.1.0`

ivtoskov:
This PR introduces a one-line change that forces the `memoize` function to respect keyword arguments when `args_to_ignore` is supplied.

- fixes #557

Checklist:

### DIFF
--- a/src/flask_caching/__init__.py
+++ b/src/flask_caching/__init__.py
@@ -638,7 +638,7 @@ class Cache:
 
         # If the function uses VAR_KEYWORD type of parameters,
         # we need to pass these further
-        kw_keys_remaining = list(kwargs.keys())
+        kw_keys_remaining = [key for key in kwargs.keys() if key not in args_to_ignore]
         arg_names = get_arg_names(f)
         args_len = len(arg_names)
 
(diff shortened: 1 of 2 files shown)


northernSage (approved):
thanks for the help @ivtoskov

PR #563, opened 2024-05-07
Update to use sha256 instead of md5

mamccorm:
Updating several references to md5, to use a more secure hash such as sha256.

Checklist:


john0isaac (commented):
This doesn't update the docstrings and tests.

PR #565, opened 2024-05-10
Handle exception on jinja template while having cache failure

### LINKED ISSUE #564: jinja caching won't have exception while having redis failure
I constantly got this error inside of jinja file
using `{% cache 900,'home' %}{% endcache %}`

`redis.exceptions.ResponseError: MISCONF Redis is configured to save RDB snapshots, but is currently not able to persist on disk. Commands that may modify the data set are disabled. Please check Redis logs for details about the error.`

It breaks the whole page if there is error in redis.

python 3.8.1
flask-caching 2.2.0

dzpt:
Due to #564 , if the debug is false in production, having cache system failure would ruin the whole app.
This is inconsistent compare to the exception handle in the `@memoize` decoration, which is allowed your app continue running even caching system (like redis) is down.

- fixes #564 

Checklist:

- [ x] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [ x] Add or update relevant docs, in the docs folder and in code.
- [ x] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.
- [ x] Add `.. versionchanged::` entries in any relevant code docs.
- [ x] Run `pre-commit` hooks and fix any issues.
- [ x] Run `pytest` and `tox`, no tests failed.

### DIFF
--- a/src/flask_caching/jinja2ext.py
+++ b/src/flask_caching/jinja2ext.py
@@ -93,11 +93,22 @@ class CacheExtension(Extension):
 
         #: Delete key if timeout is 'del'
         if timeout == "del":
-            cache.delete(key)
-            return caller()
+            try:
+                cache.delete(key)
+                return caller()
+            except Exception as e:
+                if cache.app.debug:
+                    raise e
+                else:
+                    return caller()
 
-        rv = cache.get(key)
-        if rv is None:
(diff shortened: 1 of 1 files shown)


sh4nks:
Thanks!

PR #579, opened 2024-07-14
fix - #570 - hit indication always true

### LINKED ISSUE #570: response_hit_indication hit_cache always True once set
When using a Flask route that contains URL parameters with response_hit_indication set to True, the caching functionality works however once the hit_cache header is added for the first time, it persists even if the cache isn't hit on a following request

In this example:
```
@app.route('/<param1>/<param2>', methods=['GET'])
@cache.cached(timeout=60, response_hit_indication=True)
def cached_route():
   pass
```
The caching works correctly however if you hit the endpoint three times with the following parameters, this is the outcome:
1. GET /test/test (no hit_cache header)
2. GET /test/test (hit_cache header = True)
3. **PROBLEM:** GET /test/newtest (hit_cache header = True but cache wasn't actually hit)

The last request shouldn't have the hit_cache header in it's response because nothing is correctly found in the cache for that unique hash.
I narrowed down the issue to likely the following code but I'm not an expert on Flask async:
```
if response_hit_indication:
   ...
   self.app.after_request_funcs[None].append(apply_caching)
```
The Flask documentation says the correct way to mount to after_request_funcs is to use the @after_request decorator to mount it to application once.

The implemented solution means that when the after_request_funcs are looped over to ensure_sync by Flask, it loops over request number 2 again which sets the hit_cache header to True when the cache wasn't hit for request number 3.

I essentially want to use the hit_cache header for unit testing so I can ensure that the implementation of Flask-Cache is persistent across different versions of my application.

Environment:

- Python version: 3.8.9
- Flask-Caching version: 2.3.0

idoshr:
- fixes #570 

hit indication return always true on hit indication

Checklist:

- [ x] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [x ] Add or update relevant docs, in the docs folder and in code.
- [ x] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.
- [ x] Add `.. versionchanged::` entries in any relevant code docs.
- [ x] Run `pre-commit` hooks and fix any issues.
- [ x] Run `pytest` and `tox`, no tests failed.


sh4nks:
Thanks! The implementation in #596 seemed to be a bit less complex so I went with the other implementation but I added your tests.

PR #596, opened 2024-12-19
get `found` value from request 

### LINKED ISSUE #595: response_hit_indication header always true even "found" is false
### hit_cache doesn't reflect the `found` value

When the key expires, the `found` value will be `False`, but `apply_caching` method after always give `'True'` when invoked in `after_request_funcs`

https://github.com/pallets-eco/flask-caching/blob/494d49882537a6cbcfe3cb41c4df05ae8acf60ce/src/flask_caching/__init__.py#L418

expectation: 
when `found` is False, the header will be `false` in after_request invocation

Environment:

- Python version: 3.11
- Flask-Caching version: 2.3.0

sumandari:
Store the `found` value in g and use that value in `after_request` method

- fixes #595

Checklist:

### DIFF
--- a/src/flask_caching/__init__.py
+++ b/src/flask_caching/__init__.py
@@ -422,7 +422,8 @@ class Cache:
                             response.headers["hit_cache"] = g.flask_caching_hit_cache
                         return response
 
-                    self.app.after_request_funcs[None].append(apply_caching)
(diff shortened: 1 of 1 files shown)


sh4nks:
Thanks! I'll add the tests from #579

Issue #248, opened 2021-04-18
DeprecationWarning: flask_caching.backend is deprecated


bmalbusca:
Hi, I'm receiving a deprecation Warning around of flask_caching. There is any suggesting to fix the usage of flask_caching for future?

The Warning:
```
python3.8/site-packages/flask_caching/__init__.py:237: DeprecationWarning: Using the initialization functions in flask_caching.backend is deprecated.  Use the a full path to backend classes directly.

```

The code:
```
from flask_caching import Cache
...
app.config['CACHE_TYPE'] = 'simple'
app.cache = Cache(app)

```

gergelypolonkai:
I agree the message is a bit misleading, i’ll rephrase it soon.

Meanwhile, see the [relevant section of the docs](https://flask-caching.readthedocs.io/en/latest/#set-up). In this specific case, you can simply change from the old name (`simple`) to the class name behind it (`SimpleCache`).

PR #633, opened 2026-02-13
Resolve short CACHE_TYPE names to classes instead of deprecated functions

### LINKED ISSUE #632: DeprecationWarning still emitted for short CACHE_TYPE names in v2.3.1
## Description

When using short `CACHE_TYPE` names like `"simple"`, `"redis"`, `"null"`, etc., Flask-Caching emits a `DeprecationWarning`:

```
/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/flask_caching/__init__.py:145: DeprecationWarning:

Using the initialization functions in flask_caching.backend is deprecated.  Use the a full path to backend classes directly.
```

This warning has been present since at least 2021 (see #248, which was closed without a code change). As of v2.3.1, the warning is still emitted and the initialization functions in `flask_caching.backends.__init__` are still in place.

## Steps to Reproduce

```python
from flask import Flask

app = Flask(__name__)
app.config["CACHE_TYPE"] = "simple"
cache = Cache(app)
```

Running the above with `-W all` or in a test suite that captures warnings will show the DeprecationWarning.

## Root Cause

In `_set_cache()` (`src/flask_caching/__init__.py`), when `CACHE_TYPE` has no dot, the code prepends `"flask_caching.backends."` and imports the result. If the imported object is a function (e.g., `flask_caching.backends.simple`) rather than a `BaseCache` subclass, the deprecation warning fires.

The short names resolve to wrapper functions defined in `flask_caching/backends/__init__.py` (e.g., `def simple(app, config, args, kwargs)`), which simply delegate to the class's `.factory()` method.

## Issues

1. **The deprecation has been pending for ~4 years** without the deprecated code path being removed or a migration timeline being communicated.
2. **The warning message contains a typo**: "Use **the** a full path" should be "Use a full path".
3. **The warning is confusing** for users who set `CACHE_TYPE = "SimpleCache"` thinking the short class name is the "full path" — the message doesn't clearly explain what the expected value should be.

## Suggested Resolution

One of the following approaches:

### Option A: Remove the deprecated functions
Remove the wrapper functions from `flask_caching/backends/__init__.py` and update `_set_cache()` so that short names like `"simple"` map directly to the class (e.g., `SimpleCache`) rather than the wrapper function. This would eliminate the warning while preserving backward compatibility for short names.

### Option B: Improve the warning message
If the intent is to eventually require fully-qualified class paths, at minimum:
- Fix the typo ("the a" → "a")
- Include the recommended replacement value in the warning, e.g.:
  ```
  DeprecationWarning: Using CACHE_TYPE='simple' is deprecated. 
  Use CACHE_TYPE='SimpleCache' or the full path 'flask_caching.backends.simplecache.SimpleCache' instead.
  ```

### Option C: Just remove the warning
If short names will continue to be supported indefinitely, remove the DeprecationWarning since the wrapper functions work correctly and there's no concrete plan to remove them.

## Environment

- Flask-Caching version: 2.3.1
- Python version: 3.12.12
- Flask version: 3.x

hdh7485:
## Summary

This PR fixes the long-standing `DeprecationWarning` that is emitted when users set `CACHE_TYPE` to a short name like `"simple"`, `"redis"`, `"null"`, etc.

The warning has been present since at least 2021 (see #248) and is still emitted in v2.3.1. It also contains a typo ("Use **the** a full path").

## Problem

In `_set_cache()`, when `CACHE_TYPE` contains no dot, the code prepends `"flask_caching.backends."` and uses `import_string()`. For short names like `"simple"`, this resolves to the wrapper *function* `flask_caching.backends.simple()` (defined in `backends/__init__.py`), not to the `SimpleCache` class. Since the function is not a `BaseCache` subclass, the deprecation warning fires.

## Fix

Added a `_CACHE_TYPE_ALIASES` mapping that translates old-style lowercase names to their corresponding class names:

| Short name | Class name |
|---|---|
| `null` | `NullCache` |
| `simple` | `SimpleCache` |
| `filesystem` | `FileSystemCache` |
| `redis` | `RedisCache` |
| `redissentinel` | `RedisSentinelCache` |
| `rediscluster` | `RedisClusterCache` |
| `uwsgi` | `UWSGICache` |
| `memcached` | `MemcachedCache` |
| `gaememcached` | `MemcachedCache` |
| `saslmemcached` | `SASLMemcachedCache` |
| `spreadsaslmemcached` | `SpreadSASLMemcachedCache` |

When a short name is found in this mapping, it resolves to `"flask_caching.backends.SimpleCache"` (the class) instead of `"flask_caching.backends.simple"` (the function). The class passes the `isinstance(cache_factory, type) and issubclass(cache_factory, BaseCache)` check, so no warning is emitted.

Unknown short names (e.g. custom backends without a dot) still fall through to the original `"flask_caching.backends." + import_me` path.

## Changes

- **`src/flask_caching/__init__.py`**: Added `_CACHE_TYPE_ALIASES` dict and updated `_set_cache()` to use it. Removed the now-unreachable deprecation warning code.
- **`tests/test_init.py`**: Added `test_short_names_no_deprecation_warning` to verify that all short names resolve to the correct class without emitting a `DeprecationWarning`.

## Backward Compatibility

- All existing short names continue to work exactly as before.
- Fully-qualified class paths (e.g. `"flask_caching.backends.simplecache.SimpleCache"`) continue to work.
- Custom backends using dotted paths are unaffected.

Fixes #632

### DIFF
--- a/src/flask_caching/__init__.py
+++ b/src/flask_caching/__init__.py
@@ -144,13 +144,34 @@ class Cache:
 
         self._set_cache(app, config)
 
+    #: Mapping from old-style lowercase CACHE_TYPE names to backend class names.
+    #: This allows short names like "simple" to resolve directly to the class
+    #: (e.g. SimpleCache) instead of the deprecated wrapper functions.
+    _CACHE_TYPE_ALIASES = {
+        "null": "NullCache",
+        "simple": "SimpleCache",
+        "filesystem": "FileSystemCache",
+        "redis": "RedisCache",
+        "redissentinel": "RedisSentinelCache",
+        "rediscluster": "RedisClusterCache",
+        "uwsgi": "UWSGICache",
+        "memcached": "MemcachedCache",
+        "gaememcached": "MemcachedCache",
+        "saslmemcached": "SASLMemcachedCache",
+        "spreadsaslmemcached": "SpreadSASLMemcachedCache",
+    }
+
     def _set_cache(self, app: Flask, config) -> None:
         import_me = config["CACHE_TYPE"]
         if "." not in import_me:
-            plain_name_used = True
-            import_me = "flask_caching.backends." + import_me
-        else:
-            plain_name_used = False
+            # Map old-style lowercase names to class names so they resolve
+            # to BaseCache subclasses instead of deprecated wrapper functions.
+            if import_me in self._CACHE_TYPE_ALIASES:
+                import_me = (
+                    "flask_caching.backends." + self._CACHE_TYPE_ALIASES[import_me]
+                )
+            else:
+                import_me = "flask_caching.backends." + import_me
 
         cache_factory = import_string(import_me)
         cache_args = config["CACHE_ARGS"][:]
@@ -158,14 +179,6 @@ class Cache:
 
         if isinstance(cache_factory, type) and issubclass(cache_factory, BaseCache):
             cache_factory = cache_factory.factory
-        elif plain_name_used:
-            warnings.warn(
-                "Using the initialization functions in flask_caching.backend "
-                "is deprecated.  Use the a full path to backend classes "
-                "directly.",
-                category=DeprecationWarning,
-                stacklevel=2,
-            )
 
         if config["CACHE_OPTIONS"]:
             cache_options.update(config["CACHE_OPTIONS"])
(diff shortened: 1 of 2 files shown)


sh4nks:
thanks for your PR but the old name is not supported anymore

PR #635, opened 2026-04-16
Modernize project setup

sh4nks:
This is based on Flasks setup


piradata:
When would be/is schedulled the next release of this lib?

PR #638, opened 2026-04-20
Fix memoize/cached under Python 3.14 TYPE_CHECKING annotations

### LINKED ISSUE #636: eager evaluation of annotations on python 3.14
inspect.signature() eagerly evaluates annotations in Python 3.14, breaking @memoize on functions with TYPE_CHECKING-only annotations
                                                                                                                                          
  Summary
                                                                                                                                          
  On Python 3.14, flask_caching.utils.get_function_parameters calls inspect.signature(f) with the default VALUE annotation format. That   
  format forces eager evaluation of annotations via PEP 649's __annotate__. Any function decorated with @Cache.memoize (or any decorator
  that eventually reaches get_function_parameters) whose annotations reference names defined only under if TYPE_CHECKING: — or whose      
  annotations use subscripted runtime types that aren't subscriptable — raises at decoration/first-call time.

  Reproduction

```python
  from typing import TYPE_CHECKING
  from flask import Flask

  if TYPE_CHECKING:                                                                                                                       
      from mymodule import SomeType  # TYPE_CHECKING-only, not available at runtime
                                                                                                                                          
  app = Flask(__name__)                                                
  cache = Cache(app, config={"CACHE_TYPE": "SimpleCache"})                                                                                
                                                                                                                                          
  @cache.memoize(60)
  def fetch(x: int) -> "SomeType":                                                                                                        
      ...                                                              
``` 
```                                                                                                                                         
  fetch(1)  # NameError: name 'SomeType' is not defined
                                                                                                                                          
  Stack (excerpted):                                                                                                                      
    .../flask_caching/utils.py, line 25, in get_function_parameters
      return list(inspect.signature(f).parameters.values())                                                                               
    .../inspect.py, line 3321, in signature                                                                                               
      return Signature.from_callable(...)  
    .../annotationlib.py, line 1137, in _get_dunder_annotations                                                                           
      ann = getattr(obj, "__annotations__", None)                                                                                         
    <user module>, in __annotate__                                                                                                        
      def fetch(x: int) -> "SomeType":                                                                                                    
  NameError: name 'SomeType' is not defined                                                                                               
```                                                                                                                                          
  This does not occur on Python ≤ 3.13 (annotations were lazily evaluated only by get_type_hints), nor in files using from __future__     
  import annotations (PEP 563 — annotations stay as strings).                                                                             
                                                                       
  Cause                                                                                                                                   
                                                                       
  get_function_parameters only needs parameter names, not their resolved types. But inspect.signature(f) with the default format          
  (inspect.Parameter.empty → resolves to annotationlib.Format.VALUE on 3.14) evaluates every annotation even though the caller never reads
   them.                                                                                                                                  
                                                                       
  Proposed fix

  On Python 3.14+, pass annotation_format=annotationlib.Format.FORWARDREF to inspect.signature. FORWARDREF format returns ForwardRef      
  objects for unresolvable names instead of raising, and doesn't try to subscript runtime-non-subscriptable generics. Parameter names are
  unaffected.                                                                                                                             
                                                                       
  Minimal diff for src/flask_caching/utils.py:                                                                                            
```python
  import inspect                                                                                                                          
  import sys                                                           
  import string
  from collections.abc import Callable                                                                                                    
   
  try:                                                                                                                                    
      from annotationlib import Format as _AnnotationFormat  # Python 3.14+
      _SIGNATURE_KWARGS = {"annotation_format": _AnnotationFormat.FORWARDREF}                                                             
  except ImportError:  # pragma: no cover  (< 3.14)                                                                                       
      _SIGNATURE_KWARGS = {}                                                                                                              
                                                                                                                                          
  # ...                                                                                                                                   
   
  def get_function_parameters(f: Callable) -> list:                                                                                       
      """Get function parameters"""                                    
      return list(inspect.signature(f, **_SIGNATURE_KWARGS).parameters.values())
```                                                                                                                                          
  Same treatment should be applied to get_arg_names and get_arg_default indirectly (they call get_function_parameters), and to any other  
  inspect.signature / inspect.getfullargspec call in the codebase — getfullargspec in wants_args() has the same issue; a parallel change  
  there would use the same _SIGNATURE_KWARGS-equivalent or switch to inspect.signature with FORWARDREF.                                   
                                                                       
  Why this is the right fix (vs. the workarounds)                                                                                         
   
  - Preserves API semantics: get_function_parameters already doesn't look at annotations, so dropping to FORWARDREF is a pure internal    
  change — no behavior change for users whose annotations are runtime-resolvable.
  - Doesn't require users to change their code: currently affected users must either restore from __future__ import annotations, add      
  string quotes, monkey-patch flask_caching.utils, or patch third-party generics (BaseQuerySet.__class_getitem__) to compensate.          
  - Aligned with PEP 649 / PEP 749 guidance: library authors are specifically encouraged to use FORWARDREF for introspection that doesn't
  need resolved types.                                                                                                                    
                                                                       
  Test case                                                                                                                               
                                                                       
  # tests/test_py314_type_checking_annotations.py
```python
  import sys                                                                                                                              
  from typing import TYPE_CHECKING
  import pytest                                                                                                                           
                                                                       
  if TYPE_CHECKING:                                                                                                                       
      class TypeCheckingOnly: ...                                      
                                                                                                                                          
  @pytest.mark.skipif(sys.version_info < (3, 14), reason="requires 3.14 __annotate__")                                                    
  def test_memoize_does_not_evaluate_annotations(cache):                                                                                  
      @cache.memoize(60)                                                                                                                  
      def f(x: int) -> "TypeCheckingOnly":                             
          return x * 2                                                                                                                    
      # Would NameError before the fix                                                                                                    
      assert f(3) == 6
      assert f(3) == 6  # cache hit path also goes through get_function_parameters                                                        
```                                                                                                                                          
  Environment                                                                                                                             
                                                                                                                                          
  - Python 3.14.2                                                                                                                         
  - flask-caching 2.3.1 and 2.4.0 (identical utils.py)                 
  - Reproduces on any cache backend                                                                                                       
                                                                                                                                          
  References                                                                                                                              
                                                                                                                                          
  - PEP 649 – Deferred Evaluation of Annotations Using Descriptors                                                                        
  - PEP 749 – Implementing PEP 649
  - Python 3.14 What's New: inspect.signature() gains annotation_format parameter

otistamp:
## Summary

Fixes #636.

On Python 3.14, `inspect.signature(f)` and `inspect.getfullargspec(f)` eagerly evaluate annotations via PEP 649's `__annotate__`. Any function decorated with `@Cache.memoize` or `@Cache.cached` whose annotations reference names only defined under `if TYPE_CHECKING:` raises `NameError` at decoration / first-call time.

This PR:

- Routes signature introspection in `utils.py` through a `_signature(f)` helper that passes `annotation_format=annotationlib.Format.FORWARDREF` on Python 3.14+. FORWARDREF leaves unresolvable names as `ForwardRef` objects instead of raising. `get_function_parameters` only reads parameter names, so this is a pure internal change.
- Rewrites `wants_args` to use `inspect.signature` (via `_signature`) instead of `inspect.getfullargspec`, which has the same eager-evaluation problem and no `annotation_format` parameter.
- Replaces the two remaining `inspect.getfullargspec` call sites in `__init__.py` with `get_arg_names` / `wants_args`.

## Test plan

- [x] Added `tests/test_py314_annotations.py` covering `memoize`, `cached`, and `unless` with unquoted `TYPE_CHECKING`-only annotations — these fail on `master` with 3.14 and pass with this change.
- [x] Full test suite passes on Python 3.14.4 and 3.13.12 (one pre-existing failure unrelated to this PR: `test_memoize_classfunc_repr`).
- [x] `mypy` and `ruff` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@ Version 2.4.0
 - Pass CACHE_OPTIONS as kwargs to redis_from_url. :pr:`591`
 - Pass sentinel_kwargs to redis client via CACHE_OPTIONS :pr:`626`
 - Fix ``response_hit_indication`` return True always. :pr:`579`, :pr:`596`, :issue:`595`, :issue:`570`
+- Use ``FORWARDREF`` annotation format when introspecting decorated
+  functions so that ``@memoize`` / ``@cached`` work with
+  ``TYPE_CHECKING``-only annotations on Python 3.14. :issue:`636`
 
 
 Version 2.3.1
--- a/src/flask_caching/__init__.py
+++ b/src/flask_caching/__init__.py
@@ -447,7 +447,7 @@ class Cache:
                 # Convert non-keyword arguments (which is the way
                 # `make_cache_key` expects them) to keyword arguments
                 # (the way `url_for` expects them)
-                argspec_args = inspect.getfullargspec(f).args
+                argspec_args = get_arg_names(f)
 
                 for arg_name, arg in zip(argspec_args, args, strict=False):
                     kwargs[arg_name] = arg
@@ -733,11 +733,8 @@ class Cache:
         bypass_cache = False
 
         if callable(unless):
-            argspec = inspect.getfullargspec(unless)
-            has_args = len(argspec.args) > 0 or argspec.varargs or argspec.varkw
-
             # If unless() takes args, pass them in.
-            if has_args:
+            if wants_args(unless):
                 if unless(f, *args, **kwargs) is True:
                     bypass_cache = True
             elif unless() is True:
--- a/src/flask_caching/utils.py
+++ b/src/flask_caching/utils.py
@@ -1,6 +1,15 @@
 import inspect
 import string
+import sys
 from collections.abc import Callable
+
+if sys.version_info >= (3, 14):
+    import annotationlib
+
+    def _signature(f: Callable) -> inspect.Signature:
+        return inspect.signature(f, annotation_format=annotationlib.Format.FORWARDREF)
+else:
+    _signature = inspect.signature
 
 TEMPLATE_FRAGMENT_KEY_TEMPLATE = "_template_fragment_cache_%s%s"
 # Used to remove control characters and whitespace from cache keys.
@@ -10,9 +19,10 @@ null_control = ({k: None for k in del_chars},)
 
 
 def wants_args(f: Callable) -> bool:
-    """Check if the function wants any arguments"""
-    arg_spec = inspect.getfullargspec(f)
-    return bool(arg_spec.args or arg_spec.varargs or arg_spec.varkw)
+    """Check if the function wants any positional, *args, or **kwargs arguments."""
+    return any(
+        p.kind != inspect.Parameter.KEYWORD_ONLY for p in get_function_parameters(f)
+    )
 
 
 def get_function_parameters(f: Callable) -> list:
@@ -20,7 +30,7 @@ def get_function_parameters(f: Callable) -> list:
     :param f
     :return: Parameter list of function
     """
-    return list(inspect.signature(f).parameters.values())
+    return list(_signature(f).parameters.values())
 
 
 def get_arg_names(f: Callable) -> list[str]:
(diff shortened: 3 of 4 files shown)


sh4nks:
Thanks!

PR #652, opened 2026-06-28
Don't define class-scoped fixtures as instance methods

### LINKED ISSUE #651: Test failure with pytest >= 9.1
Tests fail with pytest >= 9.1, as follows:

```
___________________ ERROR at setup of TestNullCache.test_has ___________________

fixturedef = <FixtureDef argname='make_cache' scope='class' baseid='tests/test_backend_cache.py::TestNullCache'>
request = <SubRequest 'make_cache' for <Function test_has>>

    @pytest.hookimpl(wrapper=True)
    def pytest_fixture_setup(fixturedef: FixtureDef, request) -> object | None:
        asyncio_mode = _get_asyncio_mode(request.config)
        if not _is_asyncio_fixture_function(fixturedef.func):
            if asyncio_mode == Mode.STRICT:
                # Ignore async fixtures without explicit asyncio mark in strict mode
                # This applies to pytest_trio fixtures, for example
>               return (yield)
                        ^^^^^
E               pytest.PytestRemovedIn10Warning: Class-scoped fixture defined as instance method is deprecated.
E               Instance attributes set in this fixture will NOT be visible to test methods,
E               as each test gets a new instance while the fixture runs only once per class.
E               Use @classmethod decorator and set attributes on cls instead.
E               See https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method

.venv/lib/python3.12/site-packages/pytest_asyncio/plugin.py:728: PytestRemovedIn10Warning
=========================== short test summary info ============================
ERROR tests/test_backend_cache.py::TestNullCache::test_has - pytest.PytestRemovedIn10Warning: Class-scoped fixture defined as instance m...
```

To reproduce:

```
uv lock -P pytest
uv run -m pytest tests/test_backend_cache.py
```

Environment:

- Python version: any (seen on 3.12.12, 3.13.14, and 3.14.6)
- Flask-Caching version: git main (b559c7b13e90156f6b9f19a5cdefc9690db64d64)

cjwatson:
This is deprecated as of pytest 9.1; see
https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method.

- fixes #651

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 - Docs: clarify that ``@memoize`` uses ``repr(obj)``
   (or ``__caching_id__``) for the ``self``/``cls`` identity, not Python's
   :func:`id`. :issue:`555`
+- Fix test failure with pytest 9.1. :issue:`651`
 
 
 Version 2.4.0
--- a/tests/test_backend_cache.py
+++ b/tests/test_backend_cache.py
@@ -169,7 +169,8 @@ class TestRedisCache(GenericCacheTests):
         return "werkzeug-test-case:"
 
     @pytest.fixture(scope="class", autouse=True)
-    def requirements(self, redis_server):
+    @classmethod
+    def requirements(cls, redis_server):
         pass
 
     @pytest.fixture(params=(None, False, True, gen_key_prefix))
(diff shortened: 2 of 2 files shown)


sh4nks:
Thanks!

PR #655, opened 2026-08-02
Call memoize forced_update callbacks once

### LINKED ISSUE #387: force_update callable is called twice for each call to memoized function
When setting a simple callable to the force_update parameter in memoize decorator, this callable is executed twice for each call to the cached function.

```
cache_server = Cache(app, config={"CACHE_TYPE: "SimpleCache"}

def some_test_callable():
    print("test callable executed")
    return False

@cache_server.memoize(forced_update=some_test_callable)
def some_cached_function():
    print("non-cached execution")
    return True

some_cached_function() # no cache exists, should execute
some_cached_function() # cache now exists, should simply return cached value
```

Above code should produce **exactly two** "test callable executed" messages (one for each cached function call), and one "non-cached execution" message (since the second time the function is called, the cache already exists).

What actually happens is that **two** "test callable executed" messages are printed **for each** _some_cached_function_ call - this means that _forced_update_ callable is called twice. The caching mechanism works as expected, so only one "non-cached execution" message is printed.

From my brief debugging, it seems that _force_update_ callable is indeed called twice in memoize-decorated function - once by _make_cache_key (->_memoize_version->forced_update(), line 551 in \_\_init\_\_.py) and once in the actual memoize() wrapper (line 373 in \_\_init\_\_.py).

Environment:
- Python version: 3.9.13
- Flask-Caching version: 2.0.0
- I'm using Flask and Flask-caching as part of Dash app, but that shouldn't make any difference to my knowledge.

deepakganesh78:
Fixes #387

## Reproduction

With `@cache.memoize(forced_update=callback)`, a minimal SimpleCache script that calls the memoized function twice invokes `callback` four times even though there are only two decorated function calls. The callback is evaluated once while building the memoize cache key and once again before cache lookup.

## Root cause

`decorated_function.make_cache_key(...)` calls `_memoize_version(..., forced_update=forced_update)`, which evaluates the callable. The memoize wrapper then evaluates the same callable again to decide whether to bypass the cached value.

## Fix

The memoize wrapper now evaluates `forced_update` once per decorated call and passes that boolean into version handling so the version TTL refresh behavior is preserved when forced updates occur. The public memoize `make_cache_key` no longer evaluates the `forced_update` callback as a side effect.

## Compatibility notes

This preserves the existing cache key format and forced-update refresh behavior while avoiding duplicate callback side effects. Existing callbacks still receive the same positional and keyword arguments.

## Validation

- Reproduced before the fix: minimal SimpleCache script reported `forced_calls=4 function_calls=1` and failed an assertion expecting 2 forced-update calls.
- Regression check: `python -m pytest tests\test_memoize.py::test_memoize_forced_update_parameters` fails with the source fix reverted (`assert 2 == 1`) and passes with the fix.
- `python -m pytest` -> `123 passed, 81 skipped, 4 warnings` on Python 3.13.14 / Windows. The skipped tests are backend-dependent tests for unavailable external cache/server combinations.
- `python -m ruff check src\flask_caching\__init__.py tests\test_memoize.py` -> passed.
- `pre-commit run --files src\flask_caching\__init__.py tests\test_memoize.py CHANGES.rst` -> passed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+Version 2.4.2
+-------------
+
+Unreleased
+
+- Call ``@memoize`` ``forced_update`` callbacks once per decorated function
+  call instead of once while making the key and again before cache lookup.
+  :issue:`387`
+
+
 Version 2.4.1
 -------------
 
--- a/src/flask_caching/__init__.py
+++ b/src/flask_caching/__init__.py
@@ -572,7 +572,7 @@ class Cache:
         version_data_list = list(self.cache.get_many(*fetch_keys))
         dirty = False
 
-        if (
+        if forced_update is True or (
             callable(forced_update)
             and (
                 forced_update(*(args or ()), **(kwargs or {}))
@@ -859,9 +859,7 @@ class Cache:
                     source_check = self.source_check
 
                 try:
-                    cache_key = decorated_function.make_cache_key(f, *args, **kwargs)
-
-                    if (
+                    forced_update_result = (
                         callable(forced_update)
                         and (
                             forced_update(*args, **kwargs)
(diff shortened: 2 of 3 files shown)


sh4nks:
Could you rebase your PR so I can merge it? Thanks!

PR #659, opened 2026-08-08
port fixes to cacehlib and use latest version of it

john0isaac:
- use pylibmc as main memecached client like cachelib
- add service containers to github ci/cd to actually run all tests
- remove unused dump_objects and add it's fix to cachelib
- use specific ports for databases so it doesn't conflict with default ports

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -27,6 +27,27 @@ jobs:
           - { python: "3.11" }
           - { python: "3.10" }
           - { name: "typing", tox: "typing", python: "3.14" }
+    services:
+      memcached:
+        image: memcached:1.6.41
+        ports:
+          - 11212:11211
+        # health checks
+        options: >-
+          --health-cmd "timeout 5 bash -c 'echo > /dev/tcp/127.0.0.1/11211'"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+      redis:
+        image: redis:8.6.3
+        ports:
+          - 6360:6379
+        # health checks
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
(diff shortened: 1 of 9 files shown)


src/flask_caching/backends/memcache.py
```diff
@@ -82,22 +82,6 @@ def factory(
...
-        return self._client.delete_multi(new_keys)  # type: ignore[no-any-return]
-
-    def inc(self, key: str, delta: int = 1) -> int | None:
-        key = self._normalize_key(key)
-        return self._client.incr(key, delta)  # type: ignore[no-any-return]
-
-    def dec(self, key: str, delta: int = 1) -> int | None:
-        key = self._normalize_key(key)
-        return self._client.decr(key, delta)  # type: ignore[no-any-return]
-
```
john0isaac, line 100: I have ported these fixes to cachelib so no need to override them here.
these were actual bugs there.

.github/workflows/tests.yaml
```diff
@@ -27,6 +27,27 @@ jobs:
...
+      redis:
+        image: redis:8.6.3
+        ports:
+          - 6360:6379
+        # health checks
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
john0isaac, line 50: these allows you to actually run tests over live databases in ci/cd

src/flask_caching/backends/rediscache.py
```diff
@@ -101,15 +100,6 @@ def factory(
...
 
-    def dump_object(self, value: Any) -> bytes:
-        """Dumps an object into a string for redis.  By default it serializes
-        integers as regular string and pickle dumps everything else.
-        """
-        t = type(value)
-        if isinstance(t, int):
-            return str(value).encode("ascii")
-        return b"!" + pickle.dumps(value)
-
```
john0isaac, line 112: this was dead code since cachelib has no dump_objects function / do not use it.
I ported the fix to cachelib redis serializer so inc/dec don't fail

tests/conftest.py
john0isaac, line 1: these are changes related to the added service containers in ci/cd. specific ports, skip starting databases in ci/cd since they are already running.

tests/test_backend_cache.py
```diff
@@ -22,7 +22,7 @@
     redis = None
 
 try:
-    import libmc as memcache
+    import pylibmc as memcache
```
john0isaac, line 25: I used pylibmc to match main client in cachelib we will need to do the change together to libmc or other python memecached library but I think it's best to be consistent.

tests/test_basic_app.py
```diff
@@ -77,45 +77,45 @@ def test_init_app_multi_apps(app, redis_server):
...
     }
     cache = Cache()
     cache.init_app(app, config=config)
     from flask_caching.backends.rediscache import RedisCache
 
     assert isinstance(app.extensions["cache"][cache], RedisCache)
-    rconn = app.extensions["cache"][cache]._write_client.connection_pool.get_connection(
-        "foo"
-    )
+    rconn = app.extensions["cache"][
```
john0isaac, line 87: I was getting deprecation warnings from `get_connection()`

sh4nks:
Nice, thanks!

PR #663, opened 2026-08-12
CacheLib updates and docs

john0isaac:
- Add docs-auto for live docs preview using `tox run -e docs-auto`
- revert` pylibmc` group change the original issue was in the docs not in pylibmc.
- Update the docs according to latest changes in CacheLib.
- Add new `pool_size` and `pool_blocking` for Memcached.
- Fix all docs references issues by resolving all problems flagged from running command `tox run -e docs -- --nitpick`


docs/backends.rst
```diff
@@ -1,12 +1,29 @@
...
+used directly, but you can subclass it to implement your own
+:ref:`custom cache backend <custom-cache-backends>`.
+
 
 .. _nullcache:
 
 NullCache
 `````````
 
+.. currentmodule:: flask_caching.backends.nullcache
```
john0isaac, line 23: These are needed so we can reference the classes in the docs without them referencing our classes is not resolved.

docs/backends.rst
```diff
@@ -22,9 +39,14 @@ Cache that doesn't cache
...
 ```````````
 
+.. currentmodule:: flask_caching.backends.simplecache
+.. py:class:: SimpleCache
+   :no-index-entry:
+
 Set ``CACHE_TYPE`` to ``SimpleCache`` to use this type.
 
-Uses a local python dictionary for caching. This is not really thread safe.
+Uses a local python dictionary for caching. All operations are protected by
```
john0isaac, line 48: SimpleCache is thread-safe since CacheLib 0.15.0

docs/backends.rst
```diff
@@ -175,8 +224,10 @@ Relevant configuration values
 - CACHE_MEMCACHED_USERNAME
 - CACHE_MEMCACHED_PASSWORD
 
-.. note:: Unlike MemcachedCache, SASLMemcachedCache can be configured with
-          CACHE_OPTIONS.
+.. note:: Like ``MemcachedCache``, ``pool_size``/``pool_blocking`` in
```
john0isaac, line 227: these were added for thread-safety too.

pyproject.toml
```diff
@@ -48,18 +48,19 @@ docs = [
...
+docs-auto = [
+    "sphinx-autobuild",
+]
 tests = [
     "flask[async]",
     "pytest",
     "pytest-cov",
     "pytest-asyncio",
     "pytest-xprocess",
+    "pylibmc",
```
sh4nks, line 60: I changed this because I couldnt install it on my macbook

john0isaac: I thought you did it because the docs build was failing.
Not sure how to proceed here.. ~maybe revert my change and go back to libmc like you had before?~

I recommend using [python-memcached](https://pypi.org/project/python-memcached/) not `libmc` since all of my `CacheLib` tests work for `python-memecached` but do not work for `libmc`

john0isaac: what are the errors you get when running `uv sync` to install?

sh4nks: Will post the error but it was something along the lines that pylibmc couldn't be installed due to a header not being found during compilation. It looks like that there are no wheels for this lib on mac.

john0isaac: ohh I got this error too when I tried using it the first time. I know your problem one second will get the solution.

john0isaac: You need to install `libmemcached` first from homebrew. https://stackoverflow.com/questions/14803310/error-when-install-pylibmc-using-pip/19432949#19432949

Use this command to install then run `uv sync` again.
```sh
brew install libmemcached
```

If it still can't find it (it happened with me) it's because it can't find the references to the headers even after installation then run these commands to export them and try `uv sync` again.
```sh
export CFLAGS="-I$(brew --prefix libmemcached)/include"
export LDFLAGS="-L$(brew --prefix libmemcached)/lib"
```

sh4nks (approved):

